PR #1, opened 2017-09-11
Add nodeSelector description for Migrations

rmohr:
Document that influencing the migration target via node labels is
possible.


fabiand:
@rmohr what does actually happen if there is more than one node suited (based on the labels) as a migration destination?

rmohr:
Behind the scene, a new pod is started for the VM, so a normal scheduler run is performed. Whatever the scheduler thinks is the best one out of the equals will be selected.

PR #6, opened 2017-10-26
Add a page describing installation and usage of PVMs

mpolednik:


vm-persistence.md
```diff
@@ -0,0 +1,107 @@
+## VM persistence
+
+### Preparation
+
+To persist a VM, the `Persistent Virtual Machine` add-on must be installed
+first.
+
+```bash
+git clone https://github.com/petrkotas/virt-vmconfig-crd
```
fabiand: I'd not like to reference personal repositories in the user-guide, instead I'd like to see that we get it to the incubator where we have an at least lean review of what is going in.

Thus once it's there we can happily add this to the official documentation.

mpolednik:
@fabiand agree. It might make sense to do that sooner than later, what do you think @petrkotas?

petrkotas:
Of course. The kubevirt-incubator makes total sense. Let me just finish the stuff and we can put it in there. I do not believe there should be half-baked cake in the incubator.

vm-persistence.md
```diff
@@ -0,0 +1,107 @@
...
+the actual PVM object can be created using a kubectl:
+
+```bash
+kubectl create -f example.yaml
+```
+
+#### Starting a PVM
+
+```bash
+kubectl get pvm testvm -o json | jq ".kind = 'VirtualMachine"
```
rmohr: I think that there is a part of the  command missing.

mpolednik:
First, I agree. Extra section definitely sounds like a good idea (although formatting will become painful pretty soon, unless the extra section is just unformatted collection of documents). As for the version/testing part, some kind of table makes sense -- but how do we make that not annoying to maintain?

fabiand:
According to the latest discussion, OfflineVMs will become part of the core runtime, then it can also be part of the core documentation. (Thus no separate section for this functionality).

Until then we can either close this PR or update it once implemented.

fabiand:
@mpolednik can this PR be dropped, I think a PR with similar contet (but up to date) was merged lately.

mpolednik:
@fabiand I believe you've answered the question yourself. :)

PR #9, opened 2018-01-05
Add OfflineVirtualMachine userguide

petrkotas:
This pull request add the offline virtual machine user guide.

It is still heavy work in progress and place for discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/user-guide/9)

### DIFF
--- a/offline-virtual-machine.md
+++ b/offline-virtual-machine.md
@@ -1,0 +1,74 @@
+# The Offline Virtual Machine user guide
+
+If you require to get hold of non-running virtual machines, you will need to work
+with the OfflineVirtualMachine object. The OfflineVirtualMachine holds the
+template for a running VirtualMachine and holds additional metadata used
+for tracking in the KubeVirt ecosystem.
+
+The OfflineVirtualMachine work with the VirtualMachine, which is a running
+VirtualMachine, that is created upon user request for virtual machine to start.
+
+## When to use it
+
+Whenever you need to manipulate stopped virtual machine (changing/displaying
+configuration).
+
+## How to use it
+
+The OfflineVirtualMachine is designed as a [Kubernetes CRD](https://kubernetes.io/docs/concepts/api-extension/custom-resources/).
+CRD implies that you can use the OfflineVirtualMachine as you would use any
+other Kubernetes object.
+
+### Commandline
+
+Whenever you want to manipulate the OfflineVirtualMachine through the commandline
+you can use the kubectl command. The following are examples demonstrating how
+to do it.
+
+```bash
+# Define an OfflineVirtualMachine:
+kubectl create -f myofflinevm.yaml
+
+# Start an OfflineVirtualMachine:
+kubectl patch offlinevirtualmachine myvm -p \
+    '{"spec":{"running" :"true"}}'
+
+# Look at OfflineVirtualMachine status and associated events:
+kubectl describe offlinevirtualmachine myvm
+
+# Look at the now created VirtualMachine status and associated events:
+kubectl describe virtualmachine myvm
+
+# Stop an OfflineVirtualMachine:
+kubectl patch offlinevirtualmachine myvm -p \
+    '{"spec":{"running":"false"}}'
+
+# Implicit cascade delete (first deletes the vm and then the ovm)
+kubectl delete offlinevirtualmachine myvm
+
+# Explicit cascade delete (first deletes the vm and then the ovm)
+kubectl delete offlinevirtualmachine myvm --cascade=true
+
+# Orphan delete (The running vm is only detached, not deleted)
+# Recreating the ovm would lead to the adoption of the vm
+kubectl delete offlinevirtualmachine myvm --cascade=false
+```
+
+### REST API
+
+Third party apps can utilize the Kubernetes REST API to manipulate the
+OfflineVirtualMachine.
+
+The REST API copies the structure of the Kubernetes, so it can be accessed as
+
+```text
+POST /apis/kubevirt.io/v1alpha1/namespaces/{namespace}/offlinevirtualmachine
+GET /apis/kubevirt.io/v1alpha1/namespaces/{namespace}/offlinevirtualmachine
+GET /apis/kubevirt.io/v1alpha1/namespaces/{namespace}/offlinevirtualmachine/{name}
+DELETE /apis/kubevirt.io/v1alpha1/namespaces/{namespace}/offlinevirtualmachine/{name}
+PUT /apis/kubevirt.io/v1alpha1/namespaces/{namespace}/offlinevirtualmachine/{name}
+PATCH /apis/kubevirt.io/v1alpha1/namespaces/{namespace}/offlinevirtualmachine/{name}
+```
+
+### The deletion of objects
+WIP
--- a/offline-virtual-machine.md
+++ b/offline-virtual-machine.md
@@ -10,14 +10,79 @@ VirtualMachine, that is created upon user request for virtual machine to start.
 
 ## When to use it
 
-Whenever you need to manipulate stopped virtual machine (changing/displaying
-configuration).
+Whenever you need to manipulate virtual machine (changing/displaying
+configuration) from within the Kubernetes cluster and do not want to loose
+it after it is stopped or deleted.
 
-## How to use it
+### Example
 
-The OfflineVirtualMachine is designed as a [Kubernetes CRD](https://kubernetes.io/docs/concepts/api-extension/custom-resources/).
-CRD implies that you can use the OfflineVirtualMachine as you would use any
-other Kubernetes object.
+The OfflineVirtualMachine is defined as any other Kubernetes object. Following
+is the example of OfflineVirtualMachine defined in minimal setup and
+cirros disk with cirros OS prepared.
+
+```yaml
+apiVersion: kubevirt.io/v1alpha1
+kind: OfflineVirtualMachine
+metadata:
+  name: my-vm
+spec:
+  running: false
+  selector:
+    template:
+      metadata:
+        labels:
+          anylabel: any_label
+    spec:
+      domain:
+        resources:
+          requests:
+            memory: 64M
+        devices:
+          disks:
+            - name: registrydisk
+              volumeName: registryvolume
+              disk:
+                bus: virtio
+            - name: cloudinitdisk
+              volumeName: cloudinitvolume
+              disk:
+                bus: virtio
+      volumes:
+        - name: registryvolume
+          registryDisk:
+            image: kubevirt/cirros-registry-disk-demo:devel
+```
+
+When setting the `spec.running = true` the VirtualMachine following VirtualMachine
+is created:
+
+```yaml
+apiVersion: kubevirt.io/v1alpha1
+kind: VirtualMachine
+metadata:
+  name: my-vm
+  labels:
+    anylabel: any_label
+spec:
+  domain:
+    resources:
+      requests:
+        memory: 64M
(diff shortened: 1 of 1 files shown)


offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+
+The OfflineVirtualMachine work with the VirtualMachine, which is a running
+VirtualMachine, that is created upon user request for virtual machine to start.
+
+## When to use it
+
+Whenever you need to manipulate stopped virtual machine (changing/displaying
+configuration).
+
+## How to use it
```
fabiand: This can be omitted from the user-guide, but should be in the developer docs.

offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Explicit cascade delete (first deletes the vm and then the ovm)
+kubectl delete offlinevirtualmachine myvm --cascade=true
+
+# Orphan delete (The running vm is only detached, not deleted)
+# Recreating the ovm would lead to the adoption of the vm
+kubectl delete offlinevirtualmachine myvm --cascade=false
+```
+
+### REST API
```
fabiand: REST API is not intended for the user-guide, this is to low level.

offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+kubectl describe offlinevirtualmachine myvm
+
+# Look at the now created VirtualMachine status and associated events:
+kubectl describe virtualmachine myvm
+
+# Stop an OfflineVirtualMachine:
+kubectl patch offlinevirtualmachine myvm -p \
+    '{"spec":{"running":"false"}}'
+
+# Implicit cascade delete (first deletes the vm and then the ovm)
```
stu-gott: Suggest using "OfflineVirtualMachine" and "VirtualMachine" instead of "ovm" and "vm" for consistency.

offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+# Recreating the ovm would lead to the adoption of the vm
+kubectl delete offlinevirtualmachine myvm --cascade=false
+```
+
+### REST API
+
+Third party apps can utilize the Kubernetes REST API to manipulate the
+OfflineVirtualMachine.
+
+The REST API copies the structure of the Kubernetes, so it can be accessed as
```
stu-gott: nit: remove "the" before Kubernetes since it's being referred to directly.

offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+with the OfflineVirtualMachine object. The OfflineVirtualMachine holds the
+template for a running VirtualMachine and holds additional metadata used
+for tracking in the KubeVirt ecosystem.
+
+The OfflineVirtualMachine work with the VirtualMachine, which is a running
+VirtualMachine, that is created upon user request for virtual machine to start.
+
+## When to use it
+
+Whenever you need to manipulate stopped virtual machine (changing/displaying
```
rmohr: The sentence looks incomplete.

fabiand:
ping?

petrkotas:
@rmohr thanks for the comments. Will work on it right away.

petrkotas:
Hey guys. Just FYI the minikube is putting effort to replace the KVM driver by [KVM2](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#kvm2-driver), which is maintained by the minikube team.

offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+
+Whenever you want to manipulate the OfflineVirtualMachine through the commandline
+you can use the kubectl command. The following are examples demonstrating how
+to do it.
+
+```bash
+# Define an OfflineVirtualMachine:
+kubectl create -f myofflinevm.yaml
+
+# Start an OfflineVirtualMachine:
```
pkliczewski: Will that start offline or livevm? Based on the above comments it looks like it will start an offlinevm and in my opinion we should be clear that it creates livevm.

offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+kubectl patch offlinevirtualmachine myvm -p \
+    '{"spec":{"running" :"true"}}'
+
+# Look at OfflineVirtualMachine status and associated events:
+kubectl describe offlinevirtualmachine myvm
+
+# Look at the now created VirtualMachine status and associated events:
+kubectl describe virtualmachine myvm
+
+# Stop an OfflineVirtualMachine:
```
pkliczewski: Similar comment to above. We stop livevm.

offline-virtual-machine.md
rmohr: Well it will start the OfflineVirtualMachine by creating a VirtualMachine. I think there is no difference. There  exists nothing like "start an OfflineVirtualMachine now and start the corresponding VirtualMachine later".

@pkliczewski  Maybe rewriting it to "Start an OfflineVirtualMachine by creating the VirtualMachine object" removes the ambiguity?

rmohr: Naming it `StatefulVirtualMachine` insteasd of `OfflineVirtualMachine`  would probably have made that clearer.

fabiand: Maybe this can also be made clearer if the title is user focused (not technical): "# Start the virtual machine"

offline-virtual-machine.md
fabiand: As above, maybe more user focused: `# Stop the virtual machine`

offline-virtual-machine.md
pkliczewski: @fabiand Your rewrite looks good to me

offline-virtual-machine.md
pkliczewski: @fabiand Your rewrite looks good to me

offline-virtual-machine.md
```diff
@@ -0,0 +1,74 @@
...
+you can use the kubectl command. The following are examples demonstrating how
+to do it.
+
+```bash
+# Define an OfflineVirtualMachine:
+kubectl create -f myofflinevm.yaml
+
+# Start an OfflineVirtualMachine:
+kubectl patch offlinevirtualmachine myvm -p \
+    '{"spec":{"running" :"true"}}'
```
fabiand: Please add `--type merge` and also: `"running":true}}"` - notice the boolean true

offline-virtual-machine.md
```diff
@@ -10,14 +10,79 @@ VirtualMachine, that is created upon user request for virtual machine to start.
...
+is the example of OfflineVirtualMachine defined in minimal setup and
+cirros disk with cirros OS prepared.
+
+```yaml
+apiVersion: kubevirt.io/v1alpha1
+kind: OfflineVirtualMachine
+metadata:
+  name: my-vm
+spec:
+  running: false # equivalent to replicas 0
```
fabiand: replicase can be confusing here.

offline-virtual-machine.md
```diff
@@ -10,14 +10,79 @@ VirtualMachine, that is created upon user request for virtual machine to start.
...
+        - name: registryvolume
+          registryDisk:
+            image: kubevirt/cirros-registry-disk-demo:devel
+```
 
-The OfflineVirtualMachine is designed as a [Kubernetes CRD](https://kubernetes.io/docs/concepts/api-extension/custom-resources/).
-CRD implies that you can use the OfflineVirtualMachine as you would use any
-other Kubernetes object.
+When setting the `spec.running = true` the VirtualMachine following VirtualMachine
+is created.
```
fabiand: please close wht a colon ":"

offline-virtual-machine.md
fabiand: This needs to be addressed!

offline-virtual-machine.md
```diff
@@ -39,36 +104,35 @@ kubectl describe offlinevirtualmachine myvm
 # Look at the now created VirtualMachine status and associated events:
 kubectl describe virtualmachine myvm
 
-# Stop an OfflineVirtualMachine:
-kubectl patch offlinevirtualmachine myvm -p \
+# Stop the VirtualMachine created by this OfflineVirtualMachine:
```
fabiand: Just speak of virtual machine (avoid using the kind name): "Stop the virtual machine"

As a user you just think of it as starting and stopping a VM - you don't necessarily think about the kinds.

offline-virtual-machine.md
```diff
@@ -39,36 +104,35 @@ kubectl describe offlinevirtualmachine myvm
...
+kubectl patch offlinevirtualmachine myvm --type merge -p \
     '{"spec":{"running":"false"}}'
```
fabiand: `false` instead of `"false"`

offline-virtual-machine.md
```diff
@@ -39,36 +104,35 @@ kubectl describe offlinevirtualmachine myvm
...
 
-# Implicit cascade delete (first deletes the vm and then the ovm)
+# Implicit cascade delete (first deletes the VirtualMachine and then the OfflineVirtualMachine)
 kubectl delete offlinevirtualmachine myvm
 
-# Explicit cascade delete (first deletes the vm and then the ovm)
+# Explicit cascade delete (first deletes the VirtualMAchine and then the OfflineVirtualMachine)
```
fabiand: `Ma` instead of `MA`

offline-virtual-machine.md
```diff
@@ -39,36 +104,35 @@ kubectl describe offlinevirtualmachine myvm
...
 
-# Orphan delete (The running vm is only detached, not deleted)
-# Recreating the ovm would lead to the adoption of the vm
+# Orphan delete (The running VirtualMachine is only detached, not deleted)
+# Recreating the OfflineVirtualMachine would lead to the adoption of the VirtualMachine
 kubectl delete offlinevirtualmachine myvm --cascade=false
 ```
 
-### REST API
+## What it does
```
fabiand: `How it works - The realtionship between VirtualMachine and OfflineVirtulMachine`

offline-virtual-machine.md
```diff
@@ -39,36 +104,35 @@ kubectl describe offlinevirtualmachine myvm
...
-PATCH /apis/kubevirt.io/v1alpha1/namespaces/{namespace}/offlinevirtualmachine/{name}
-```
+When set `spec.running = true` the OfflineVirtualMachine will try to keep the
+VirtualMachine running. This means: If the VirtualMachine is turned off from
+within the guest or fails. It will recreate it and start it again. The only
+way to turn the VirtualMachine off is by setting `spec.running = false`.
 
-### The deletion of objects
-WIP
+OfflineVirtualMachine also propagates the VirtualMachine spec stored in
```
fabiand: "The OfflineVirtualMachine template works just like Pod templates: Like a cookie cutter. This implies that the spec is only used for creating the new VM. Once a VM was created, then it's spec will not be changed."

Or something along those lines.

The first sentence somehow looks a little like it could be _propgagted_ at runtime.

petrkotas:
@fabiand Addressed.

offline-virtual-machine.md
```diff
@@ -0,0 +1,140 @@
...
+you can use the kubectl command. The following are examples demonstrating how
+to do it.
+
+```bash
+# Define an offline virtual machine:
+kubectl create -f myofflinevm.yaml
+
+# Start the virtual machine:
+kubectl patch offlinevirtualmachine myvm --type merge -p \
+    '{"spec":{"running" :true}}'
```
fabiand: the space also needs to be removed - the space before ":"

PR #11, opened 2018-01-05
Add VirtualMachinePreset user guide

stu-gott:
VirtualMachinePreset user guide

### DIFF
--- a/vm-presets.md
+++ b/vm-presets.md
@@ -17,15 +17,14 @@ KubeVirt uses Kubernetes `Labels` and `Selectors` to determine which
 Annotation upon successful completion.
 
 Any domain structure can be listed in the `spec` of a `VirtualMachinePreset`.
-e.g. Clock, Features, Memory, CPU, or Devices such network interfaces or disks.
-All elements of the `spec` section of a `VirtualMachinePreset` will be applied
+e.g. Clock, Features, Memory, CPU, or Devices such network interfaces.  All
+elements of the `spec` section of a `VirtualMachinePreset` will be applied
 to the `VirtualMachine`.
 
 Once a `VirtualMachinePreset` is successfully applied to a `VirtualMachine`,
 the `VirtualMachine` will be marked with an annotation to indicate that it
-was applied. If an error occurs while a `VirtualMachinePreset` is being applied
-(for example, if a conflict occurs), none of the `VirtualMachinePreset` will be
-applied.
+was applied. If a conflict occurs while a `VirtualMachinePreset` is being
+applied that portion of the `VirtualMachinePreset` will be skipped.
 
 
 Conflicts
@@ -34,28 +33,26 @@ Conflicts
 `VirtualMachinePresets` use a similar conflict resolution strategy to
 Kubernetes `PodPresets`. If a portion of the domain spec is present in both a
 `VirtualMachine` and a `VirtualMachinePreset` and both resources have the
-identical information, then no error will occur and `VirtualMachine` creation
+identical information, then no conflict will occur and `VirtualMachine` creation
 will continue normally. If however there is a conflict between the resources,
-an error will occur and the `VirtualMachine` will not be created. For example:
-If both the `VirtualMachine` and `VirtualMachinePreset` define a `Volume`, but
-use different paths for the same name, KubeVirt will note the conflict.
+an Event will be created indicating which `DomainSpec` element of which
+`VirtualMachinePreset` was problematic. For example: If both the `VirtualMachine`
+and `VirtualMachinePreset` define a `CPU`, but use a different number of `Cores`,
+KubeVirt will note the conflict.
+
+If any settings from the `VirtualMachinePreset` were successfully applied, the
+`VirtualMachine` will still be annotated.
 
 In the event that a conflict occurs, KubeVirt will create an `Event`.
-`kubectl describe` can be used to show all `Events`. For example:
+`kubectl get events` can be used to show all `Events`. For example:
 
 ```
-$ kubectl describe ...
+$ kubectl get events
 ....
 Events:
   FirstSeen                         LastSeen                        Count From                              SubobjectPath                Reason    Message
-  Thu, 11 Jan 2018 08:34:14 -0500   Thu, 11 Jan 2018 08:34:14 -0500 1     {vmpreset.admission.kubernetes.io/vmpreset-allow-database }    conflict  Conflict on virtual machine preset. Duplicate mountPath /cache.
-
+  2m          2m           1         myvm.1515bbb8d397f258                       VirtualMachine                                     Warning   Conflict                  virtualmachine-preset-controller   Unable to apply VirtualMachinePreset 'example-preset': spec.cpu: &{6} != &{4}
 ```
-
-Because `VirtualMachinePresets` are implemented as an Initializer within the
-`virt-controller` pod, log messages associated with resource conflicts will
-also be reflected there.
-
 
 Creation and Usage
 ------------------------
@@ -69,12 +66,13 @@ KubeVirt will determine which `VirtualMachinePresets` apply to a Particular
 `VirtualMachine` by matching `Labels`. For example:
 
 ```yaml
+apiVersion: kubevirt.io/v1alpha1
 kind: VirtualMachinePreset
 metadata:
   name: example-preset
   selector:
     matchLabels:
-      flavor: foo
+      kubevirt.io/flavor: foo
   ...
 ```
 
@@ -82,15 +80,15 @@ would match any `VirtualMachine` in the same namespace with a `Label` of
 `flavor: foo`. For example:
 
 ```yaml
+apiVersion: kubevirt.io/v1alpha1
 kind: VirtualMachine
 version: v1
 metadata:
   name: myvm
   labels:
-    flavor: foo
+    kubevirt.io/flavor: foo
   ...
 ```
-
 
 Examples
 =============================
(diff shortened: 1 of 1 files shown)


fabiand:
This rather looks like a design doc which would fit into `kubevirt/docs`. What do you think?

vm-presets.md
```diff
@@ -0,0 +1,185 @@
...
+------------------------
+
+`VirtualMachinePresets` are implemented as a Kubernetes `Initializer`. This
+means the `VirtualMachine` resource is modified before it is visible to or
+processed by any other component of KubeVirt.
+
+One noteworthy difference between the behavior of a `VirtualMachinePreset` and
+a `PodPreset` is that if an error occurs during processing of a `PodPreset`,
+Kubernetes will continue to create the `Pod` without applying the `PodPreset`.
+KubeVirt will move the VirtualMachine to the `Failed` state instead. The
```
rmohr: As you write it ourself the vm to Failed is something different than what the PodPreset does with a Pod. I would expect that we don't add the VirtualMachinePreset annotation to the vm, log an error event which we can see with `kubectl describe vm myvm` and continue with the next preset.

vm-presets.md
```diff
@@ -0,0 +1,185 @@
...
+rationale behind this difference is that settings defined in a
+`VirtualMachinePreset` are presumed to be critical to the function of the
```
rmohr: I am not sure I can follow that rationale. What the VirtualMachinePreset applies, may or may not be critical for an Operating System inside the vm, but it should never be critical to launch a VirtualMachine.

Here one example flow where this is problematic:

 1. Create a vm
 2. Presets exand it
 3. I fetch the running vm spec
 4. I delete my vm
 5. Admin changes the presets
 6. I want to post the fetched spec again
 7. I can't because of a Preset merge conflict.

vm-presets.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+```yaml
+kind: VirtualMachinePreset
+metadata:
+  name: windows-10
+  selector:
+    matchLabels:
+      flavor: windows-10
+spec:
```
rmohr: Could you synchornize the Preset examples with the current api? I think you see most of what you need in the windows-demo.yaml in the core repo.

vm-presets.md
```diff
@@ -0,0 +1,185 @@
...
+`VirtualMachine` will be put into a `Failed` state, as noted above).
+
+Once a `VirtualMachinePreset` is successfully applied to a `VirtualMachine`,
+the `VirtualMachine` will be marked with an annotation to indicate that is
+was applied.
+
+Usage
+------------------------
+
+Any domain structure can be listed in the spec of a `VirtualMachinePreset`.
```
rmohr: It might be worth adding that a preset can add elements into lists (lists inside devices and volumes for instance).

vm-presets.md
```diff
@@ -0,0 +1,185 @@
...
+uses it:
+
+
+```yaml
+kind: VirtualMachinePreset
+metadata:
+  name: windows-10
+  selector:
+    matchLabels:
+      flavor: windows-10
```
rmohr: I think what we see pretty nice here, is that we probably want a set of VirtualMachinePresets which would implement such a `flavor`. E.g. one Preset for windows features, one for clocks, one for memory and one for cpu. That would make it easy for people to override specific sections, but still don't have to worry about the whole spec.

rmohr:
Having an example on how to add your own preset to the cluster might make sense.

rmohr:
> This rather looks like a design doc which would fit into kubevirt/docs. What do you think?

Hard to tell. Maybe we can make it a little bit more example focused like https://kubernetes.io/docs/tasks/inject-data-application/podpreset/#conflict-example. In general I think it mostly contains information which I would need in order to understand how to use the VirtualMachinePresets.

vm-presets.md
stu-gott: Done!

vm-presets.md
stu-gott: Works for me.

vm-presets.md
stu-gott: Done, thanks! I missed that example.

vm-presets.md
stu-gott: I agree. It makes it easier to see the moving parts for demonstration purposes if the units are smaller. done.

vm-presets.md
```diff
@@ -0,0 +1,237 @@
+Virtual Machine Presets
+=============================
+
+`VirtualMachinePresets` are an extension to general `VirtualMachine`
+Configuration based heavily on `PodPresets` from Kubernetes. When a
```
fabiand: lower case c for configuration?
Also s/based/behaving/? Base could imply that they are technically based on the same src.

vm-presets.md
```diff
@@ -0,0 +1,237 @@
...
+`VirtualMachine` is created, any applicable `VirtualMachinePresets`
+will be applied to the existing spec for the `VirtualMachine`. This allows
+for re-use of common settings that should apply to multiple `VirtualMachines`.
+
+Usage
+------------------------
+
+KubeVirt uses Kubernetes `Labels` to determine which `VirtualMachinePresets`
```
fabiand: Labels and Selectors.

vm-presets.md
```diff
@@ -0,0 +1,237 @@
...
+apply to any given `VirtualMachine`, similarly to how `PodPresets` work in
+Kubernetes. The `VirtualMachine` is marked with an Annotation upon successful
+completion.
+
+```yaml
```
fabiand: Please add some context to this example.

Is this before or after a preset?

If after, then pls add the expected annotation.

vm-presets.md
```diff
@@ -0,0 +1,237 @@
...
+```
+
+Any domain structure can be listed in the `spec` of a `VirtualMachinePreset`.
+e.g. Clock, Features, Memory, CPU, or Devices such network interfaces or disks.
+All elements of the `spec` section of a `VirtualMachinePreset` will be applied
+to the `VirtualMachine`. Single element structures will be applied individually
+while members of lists will be appended. See the `Examples` section for more
+detail.
+
+
```
fabiand: Remove this line

vm-presets.md
```diff
@@ -0,0 +1,237 @@
...
+Once a `VirtualMachinePreset` is successfully applied to a `VirtualMachine`,
+the `VirtualMachine` will be marked with an annotation to indicate that it
+was applied. If an error occurs while a `VirtualMachinePreset` is being applied
+(for example, if a conflict occurs), none of the `VirtualMachinePreset` will be
+applied.
```
fabiand: How and where are errors shown? I.e. in case thats oemthing can not be applied?

rmohr: +1 on adding that. I would expect the same like in https://kubernetes.io/docs/tasks/inject-data-application/podpreset/#conflict-example. So when I do a `describe`, I will see an event for every merge conflict.

stu-gott:
Good idea.

vm-presets.md
stu-gott: An example covering this exact same interaction occurs later in the document, I think I'll remove it here because I agree that it conflicts with the text.

vm-presets.md
fabiand: Sounds good, we need to make sure to cover this in teh api design doc as well.

Here we can be more reöaxed and provide an example and reference PodPresets as Roman suggests.

vm-presets.md
```diff
@@ -0,0 +1,293 @@
...
+resulting resource would look like this:
+
+
+```yaml
+kind: VirtualMachine
+version: v1
+metadata:
+  name: myvm
+  labels:
+    flavor: windows-10
```
yanirq: I'm  a bit confused,   where did the "windows-10" flavor came from ? 
above i can see     flavor: default-features  both in the preset and the VM.

stu-gott: You are correct. They should match. Alas, the dangers of creating examples by hand.

fabiand:
This does not impact this work, but just to estblish the link to that discussion: As part of #731 it is being discussed what specific label should be used on the VM and for the label selector.

Once that has settled we can update the docs.

/cc @yanirq

stu-gott:
From a technical point of view, literally any set of rules we come up with is interchangeable with any other. But you're absolutely right. The getting started document on the subject should follow best practices guidelines.

yanirq:
@stu-gott   
I think we can already change the examples to use the new guest-os label
from https://github.com/kubevirt/user-guide/pull/25

also might be worth adding a short description about it (kubevirt.io/os-short-id)
other option is to wait for this user guide to get merged and then add the info above.

stu-gott:
@yanirq So to clarify to be sure I understand what you asked: we have agreed to use `kubevirt.io/os-short-id: <output from libosinfo>` as opposed to `kubvirt.io/flavor: user-defined`?

I feel like that can be artificially limiting as it presumes the reason for and distinction between different presents is exclusively tied to the operating system being targeted.

fabiand:
In the examples we settle on libosinfo. And we'll provide presets for these identifiers.

In general presets are just limited by th eimagination of the users - and what they use as label names.

stu-gott:
Ah! So it's a convention we've agreed to for the sake of example. That works. Thanks for clarifying!

yanirq:
> @yanirq So to clarify to be sure I understand what you asked: we have agreed to use kubevirt.io/os-short-id: <output from libosinfo> as opposed to kubevirt.io/flavor: user-defined?

@stu-gott   I think that at least for the examples using an OS preset we should follow the convention of 
``kubevirt.io/os-short-id:``
the supported table is currently in PR : https://github.com/kubevirt/user-guide/pull/25

vm-presets.md
```diff
@@ -0,0 +1,256 @@
...
+`VirtualMachine` is created, any applicable `VirtualMachinePresets`
+will be applied to the existing spec for the `VirtualMachine`. This allows
+for re-use of common settings that should apply to multiple `VirtualMachines`.
+
+
+Usage
+------------------------
+
+KubeVirt uses Kubernetes `Labels` and `Selectors` to determine which
+`VirtualMachinePresets` apply to any given `VirtualMachine`, similarly to how
```
rmohr: @stu-gott I think that you now implemented it differently (you also partially apply the preset).

PR #13, opened 2018-01-16
Document volumes and disks

rmohr:
Document which volume and disk types are supported in KubeVirt and how connecting them works.


disks-and-volumes.md
```diff
@@ -0,0 +1,315 @@
...
+A disk can be made accessible via four different types:
+
+ * disk
+ * cdrom
+ * floppy
+ * lun
+
+### lun
+
+A `lun` disk will expose the volume as a LUN device to the vm.
```
fabiand: Could you link to a definition of `lun` - especially in contrast to disks.

disks-and-volumes.md
```diff
@@ -0,0 +1,315 @@
...
+  domain:
+    resources:
+      requests:
+        memory: 64M
+    devices:
+      disks:
+      - name: mypvcdisk
+        volumeName: mypvc
+        # This makes it a lun device
+        lun: {}
```
fabiand: I like the empty structs for - empty structs :)

(lun: {} vs lun:)

I'll see that I have an eye on this in examples in general.

disks-and-volumes.md
```diff
@@ -0,0 +1,315 @@
...
+Supported volume sources are
+
+ * cloudInitNoCloud
+ * iscsi
+ * persistentVolumeClaim
+ * registryDisk
+
+### cloudInitNoCloud
+
+Allows attaching `cloudInitNoCloud` data-sources to the vm. If the vm contains
```
fabiand: vm vs VM?

disks-and-volumes.md
```diff
@@ -0,0 +1,315 @@
+# Volumes and Disks
+
```
fabiand: Maybe a few introductional words to state that th eterm disk coins the guest facing side, whereas volumes specifies the kube source side.

disks-and-volumes.md
rmohr: Not sure where to link. The difference is what you see in the guest. Do you mean linking to the api-reference?

disks-and-volumes.md
rmohr: `lun:` is equivalent to `lun: null`, whereas `lun: {}` is an empty struct/dict. So having `{}` means it is present, whereas `lun:` is the same as not specifying it.

disks-and-volumes.md
fabiand: IIUIC then the diffrerence between lun and disk is that with lun more scsi calls are passed through disk is somewhat filtering it - that's at least what I get from the libvirt docs.
Is there any other difference I am missing?

rmohr: I can add to the api-reference docs that this allows generic iscsi command passthrough.

fabiand: Yep, that would work for me, or even just a sentence here differentiating lun from disk. At least some hint how they differ :)

disks-and-volumes.md
rmohr: done

disks-and-volumes.md
rmohr: done

PR #18, opened 2018-01-31
machine: add section about machine type

mpolednik:
To help users understand various machine types and ways of configuring
them, we add a simple section (currently only for x86_64 arch)
explaining what machine types are, which machine type does KubeVirt
default to and how to override that default.

Signed-off-by: Martin Polednik <mpolednik@redhat.com>

### DIFF
--- a/machine.md
+++ b/machine.md
@@ -4,6 +4,41 @@ Finetuning different aspects of the hardware which are not device related
 (BIOS, mainboard, ...) is sometimes necessary to allow Guest Operating Systems
 to properly boot and reboot.
 
+## Machine Type
+
+QEMU is able to work with two different classes of chipsets for x86\_64, so
+called machine types. The x86\_64 chipsets are i440fx (also called pc) and q35.
+They are versioned based on qemu-system-$ARCH, following the format
+`pc-${machine_type}-${qemu_version}`, e.g. `pc-i440fx-2.10` and `pc-q35-2.10`.
+
+KubeVirt defaults to QEMU's newest q35 machine type. If a custom machine type is
+desired, it is configurable via following structure:
+
+```
+metadata:
+  name: myvm
+spec:
+  domain:
+    machine:
+      # This value indicates QEMU machine type.
+      type: pc-q35-2.10
+    resources:
+      requests:
+        memory: 512M
+    devices:
+      disks:
+      - name: myimage
+        volumeName: myimage
+        disk: {}
+  volumes:
+    - name: myimage
+      persistentVolumeClaim:
+        claimName: myclaim
+```
+
+Comparision of the machine types' internals can be found
+[at QEMU wiki](https://wiki.qemu.org/Features/Q35).
+
 ## BIOS/UEFI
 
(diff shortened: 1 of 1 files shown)


fabiand:
Adding some native speakers.

I have th eimpression that the introductional paragraph is a little technical - but maybe that is just me.

machine.md
```diff
@@ -4,6 +4,41 @@ Finetuning different aspects of the hardware which are not device related
 (BIOS, mainboard, ...) is sometimes necessary to allow Guest Operating Systems
 to properly boot and reboot.
 
+## Machine Type
+
+QEMU is able to work with two different classes of chipsets for x86\_64, so
```
stu-gott: Is this shorthand for "x86 and x86_64"? I haven't seen that before.

machine.md
```diff
@@ -4,6 +4,41 @@ Finetuning different aspects of the hardware which are not device related
...
+      - name: myimage
+        volumeName: myimage
+        disk: {}
+  volumes:
+    - name: myimage
+      persistentVolumeClaim:
+        claimName: myclaim
+```
+
+Comparision of the machine types' internals can be found
```
stu-gott: nit: `types'` is the plural possessive form. I think `type's` would be correct here.

machine.md
```diff
@@ -4,6 +4,41 @@ Finetuning different aspects of the hardware which are not device related
...
 to properly boot and reboot.
 
+## Machine Type
+
+QEMU is able to work with two different classes of chipsets for x86\_64, so
+called machine types. The x86\_64 chipsets are i440fx (also called pc) and q35.
+They are versioned based on qemu-system-$ARCH, following the format
+`pc-${machine_type}-${qemu_version}`, e.g. `pc-i440fx-2.10` and `pc-q35-2.10`.
+
+KubeVirt defaults to QEMU's newest q35 machine type. If custom machine type is
```
stu-gott: Nit: "If a custom machine type is desired"

machine.md
mpolednik: No, it should be markdown formatting thingy. Will check if it's needed.

machine.md
mpolednik: (just making sure) isn't plural possessive the correct choice here? It is link to docs about the internals of 2 distinct machine types.

machine.md
mpolednik: http://johnmacfarlane.net/babelmark2/?text=my%5C_name%5C_%0A%0Amy_name_ this example from SO indicates that it is the most supported underscore formatting available; in the final processed document only underscore itself is displayed.

machine.md
stu-gott: You are correct! I had to reread the statement a few times to understand your question. I had read "machine" as the object of the preposition, but as you pointed out, the comparison is between the internals of multiple types.

machine.md
rmohr: so lets fix this one type and then merge this.

mpolednik:
fixed & rebased (the SMBIOS/OVMF sections are kept together)

PR #19, opened 2018-01-31
Document that all vms are booted with BIOS

rmohr:
Document that all vms are booted with BIOS and that UEFI is not yet supported.

Signed-off-by: Travis CI <travis@travis-ci.org>


rmohr:
Not sure if it is the best location, but BIOS/UEFI is related to how the machine looks like, therefore I put it to that location.

mpolednik:
Maybe worth noting that OVMF == UEFI for VMs.

rmohr:
> Maybe worth noting that OVMF == UEFI for VMs.

done

fabiand:
Place looks fine for me, as it's part of the VM definition.

PR #22, opened 2018-02-06
OpenShift section is added to installation.md

mareklibra:
Instructions for adding SCCs are listed.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -76,6 +76,26 @@ Or to have convenience commands for:
 The most recent version of the tool can be retrieved from the [official release
 page](https://github.com/kubevirt/kubevirt/releases).
 
+## Deploying on OpenShift
+Following [SCCs](https://docs.openshift.com/container-platform/3.7/admin_guide/manage_scc.html) need to be added prior `kubevirt.yaml` deployment:
+
+```
+oc adm policy add-scc-to-user privileged system:serviceaccount:kube-system:kubevirt-privileged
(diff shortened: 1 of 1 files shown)


installation.md
```diff
@@ -76,6 +76,27 @@ Or to have convenience commands for:
...
+oc create serviceaccount -n kube-system kubevirt
+```
+
+and following [SCCs](https://docs.openshift.com/container-platform/3.7/admin_guide/manage_scc.html) needs to be added:
+```
+oc adm policy add-scc-to-user privileged system:serviceaccount:kube-system:kubevirt-privileged
+oc adm policy add-scc-to-user privileged system:serviceaccount:kube-system:kubevirt-controller
+```
+
+Please note, the `kubevirt` service account is created even within `kubevirt.yaml` but creating it explicitly allows adding SCCs before `kubevirt.yaml` is deployed.
```
fabiand: How does this make a difference?

installation.md
```diff
@@ -76,6 +76,27 @@ Or to have convenience commands for:
...
+
+### Kubevirt 0.2.0
```
fabiand: Please move this to the beginnig of the section so readers read it before deployment
,a lso maybe use bold \*\*NOTE\*\*:

installation.md
mareklibra: oh, right. No difference, declarative approach.

installation.md
mareklibra: done

fabiand:
Thanks!

PR #23, opened 2018-02-14
disks: document the 'Bus' attribute for disks

fromanirh:


disks-and-volumes.md
```diff
@@ -21,6 +21,10 @@ A disk can be made accessible via four different types:
 All possible configuration options are available in the
 [Disk API Reference](https://kubevirt.github.io/api-reference/master/definitions.html#_v1_disk).
 
+All types but **floppy** allow you to specify the `Bus` attribute. The `Bus` attribute determine
+how the disk will be presented to the Guest Operating System. **floppy** disks don't support
+the `Bus` attribute: they are always attached to the `fdc` bus.
```
fabiand: Is fdc a bus or a device name?

I suppose that bus would be IDE, not?

disks-and-volumes.md
```diff
@@ -73,7 +77,10 @@ spec:
       - name: mypvcdisk
         volumeName: mypvc
         # This makes it a disk
-        disk: {}
```
fabiand: Is there still an example with an empty disk section?

Would be good to keep this example.

disks-and-volumes.md
```diff
@@ -21,6 +21,10 @@ A disk can be made accessible via four different types:
 All possible configuration options are available in the
 [Disk API Reference](https://kubevirt.github.io/api-reference/master/definitions.html#_v1_disk).
 
+All types but **floppy** allow you to specify the `Bus` attribute. The `Bus` attribute determine
```
rmohr: could you use `bus` instead of `Bus`? The user works with json/yaml and not with the struct names.

rmohr: s/determine/determines/

disks-and-volumes.md
rmohr: +1

disks-and-volumes.md
fromanirh: fdc is a bus name (related doc: https://libvirt.org/formatdomain.html#elementsControllers)

floppies have their own bus and controller, AFAIK different from IDE. Se also https://en.wikipedia.org/wiki/Floppy-disk_controller and c

fabiand: Wow - Thank you, I was never aware of this!

fabiand:
+1 lgtm!

PR #25, opened 2018-02-14
Guest OS info label for VM

yanirq:
Added Guest OS info label for Virtual Machine to user guide.

Signed-off-by: Yanir Quinn yquinn@redhat.com

### DIFF
--- a/guest-os-info.md
+++ b/guest-os-info.md
@@ -1,0 +1,26 @@
+## Guest Operating System Information
+
+Guest operating system identity for the Virtual Machine will be provided by the label ``kubevirt.io/os`` :
+
+```
+metadata:
+  name: myvm
+  labels:
+    kubevirt.io/os: win2k12r2
+```
+
+The ``kubevirt.io/os`` label is based on the short OS identifier from [libosinfo](https://libosinfo.org/)
+database.
+The following Short IDs are currently supported:
+
+| Short ID | Name | Version | Family | ID |
+| --- | --- | --- | --- | --- |
+| **win2k12r2** | Microsoft Windows Server 2012 R2 | 6.3 | winnt | http://microsoft.com/win/2k12r2 |
+
+
+### Use with presets
+
+A Virtual Machine Preset representing an operating system with a ``kubevirt.io/os`` label could be applied on any given
+Virtual Machine that have and match the``kubevirt.io/os`` label.
+
+
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -11,6 +11,7 @@
     * [Access](access.md)
     * [Disks and Volumes](disks-and-volumes.md)
     * [Startup Scripts](startup-scripts.md)
+    * [Guest Operating System Information](guest-os-info.md)
     * [Virtualized Hardware Configuration](machine.md)
     * [Assigning VMs to Nodes](assign-vm-node.md)
   * [Controllers](workloads/controllers.md)


guest-os-info.md
```diff
@@ -0,0 +1,35 @@
...
+Guest operating system identity for the Virtual Machine will be provided by the label ``os-short-id`` :
+
+```
+metadata:
+  name: myvm
+  labels:
+    os-short-id:
+```
+
+The os-short-id label is based on the short OS identifier from [libosinfo](https://libosinfo.org/).
```
fabiand: `os-short-id`

guest-os-info.md
```diff
@@ -0,0 +1,35 @@
+## Guest Operating System Information
+
+Guest operating system identity for the Virtual Machine will be provided by the label ``os-short-id`` :
+
+```
+metadata:
+  name: myvm
+  labels:
+    os-short-id:
```
fabiand: I'd add an example value here - i.e. `win2k12r2` - and then use the example value in the `osinfo query` line below (to make the connection to the user).

guest-os-info.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+The os-short-id label is based on the short OS identifier from [libosinfo](https://libosinfo.org/).
+Libosinfo database can be queried to extract additional information about the operating system using the short-id:
+
+```
+# List all operating systems
+$ osinfo-query os
```
fabiand: here you would then say

`osinfo-query os short-id=win2k12r2`

guest-os-info.md
```diff
@@ -0,0 +1,35 @@
...
+ Short ID             | Name                            | Version | ID
+---------------------------------------------------------------------------------------------------
+...
+ fedora24             | Fedora 24                    | 24        | http://fedoraproject.org/fedora/24
+ fedora25             | Fedora 25                    | 25        | http://fedoraproject.org/fedora/25
+ fedora26             | Fedora 26                    | 26        | http://fedoraproject.org/fedora/26
+...
+
+```
+
```
fabiand: You might want to add that there are additional fields to query, man osinfo-query will help to find them.

yabe alo a note that there are xml files with the same informations? but maybe this is for the developer guide.

guest-os-info.md
```diff
@@ -0,0 +1,35 @@
...
+### Use with presets
+
+A Virtual Machine Preset representing an operating system could be applied on any given Virtual Machine using the
+os-short-id label.
```
fabiand: `os-short-id`

guest-os-info.md
```diff
@@ -0,0 +1,35 @@
...
+ fedora24             | Fedora 24                    | 24        | http://fedoraproject.org/fedora/24
+ fedora25             | Fedora 25                    | 25        | http://fedoraproject.org/fedora/25
+ fedora26             | Fedora 26                    | 26        | http://fedoraproject.org/fedora/26
+...
+
+```
+
+### Use with presets
+
+A Virtual Machine Preset representing an operating system could be applied on any given Virtual Machine using the
```
fabiand: > A Virtual machine Preset suited for a specific guest can then have a label selector to match a present against the `os-short-id` label of a VM.

Along those lines. I'd be explicit thata the preset should math against the os-short-id label.

guest-os-info.md
```diff
@@ -0,0 +1,49 @@
...
+
+ Short ID             | Name                                              | Version | ID
+----------------------------------------------------------------------------------------------------------------------
+...
+ fedora24             | Fedora 24                                      | 24        | http://fedoraproject.org/fedora/24
+ fedora25             | Fedora 25                                      | 25        | http://fedoraproject.org/fedora/25
+ fedora26             | Fedora 26                                      | 26        | http://fedoraproject.org/fedora/26
+...
+ win2k12r2          | Microsoft Windows Server 2012 R2 | 6.3       | http://microsoft.com/win/2k12r2
+...
```
pkliczewski: If you want to have a table with short IDs please make sure it is complete

guest-os-info.md
```diff
@@ -0,0 +1,49 @@
...
+# Get OS specific info
+$ osinfo-query os short-id=win2k12r2
+
+ Short ID             | Name                                              | Version | ID
+----------------------------------------------------------------------------------------------------------------------
+ win2k12r2          | Microsoft Windows Server 2012 R2 | 6.3       | http://microsoft.com/win/2k12r2
+
+```
+
+Addtional properties can be queried include : name, version, family, vendor, release-date, eol-date, codename, id.
```
pkliczewski: You say about ability to query. How one could do it? Please make sure to include this info here.

guest-os-info.md
yanirq: The table here is a shorten version of the ``osinfo-query os`` output (that include all short IDs)
I didnt see a need to show the whole table here

pkliczewski: In my opinion if you create something partial like above you either finalize it or give a reference where one can find complete list of `short-id`s.

guest-os-info.md
```diff
@@ -0,0 +1,53 @@
+## Guest Operating System Information
+
+Guest operating system identity for the Virtual Machine will be provided by the label ``os-short-id`` :
+
+```
+metadata:
+  name: myvm
+  labels:
+    os-short-id: win2k12r2
```
rmohr: We need to prefix it with `kubevirt.io/`, to make clear that it belongs to us.

guest-os-info.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+The ``os-short-id`` label is based on the short OS identifier from [libosinfo](https://libosinfo.org/).
```
rmohr: Do we really want to expose `kubevirt.io/os-id` too? I would prefer not to, since the not-short is is a mix of a http url + an identifier. Also how would we handle conflicting identifiers then?

Consider renaming the label to `kubevirt.io/os`.

yanirq: We want to expose only ``kubevirt.io/os-short-id``   (or ``kubevirt.io/os`` if we decide to rename it)
using only the os-short-id we expect that other properties could be extracted from libosinfo if one desire to do so. 

If we use only the os-short-id  it should be unique , or do you refer to other type of conflicts ?

fabiand: +1 on exposing `kubevirt.io/os-short-id` only

fabiand:
@rmohr To me it look slike it can be quite valuable to state that we align on the libosinfo short-ids.

Because it's important that users and developers know that they can use that database to extract further informations.

rmohr:
> @rmohr To me it look slike it can be quite valuable to state that we align on the libosinfo short-ids.
> 
> Because it's important that users and developers know that they can use that database to extract further informations.

Sure but one sentence about it as a side note is good enough. I don't expect that the user says: "ah of course libosinfo, now everything is clear to me, that explains everything, I know that tool and that dabase by hear".

In my opinion it is by far more important to have an explicit table of presets which we provide and to which OS they map.

fabiand:
Right. I agree about the technical depth.

I'd then suggest to add a few more informations to the kubevirt core repository with some developer focused informations.

yanirq:
@fabiand @rmohr   
Narrowed down the information for users with a full supported table.
additional dev information created in another pull request :  https://github.com/kubevirt/kubevirt/pull/767

guest-os-info.md
```diff
@@ -0,0 +1,34 @@
...
+| **rhel7.3** | Red Hat Enterprise Linux 7.3 | 7.3 | linux | http://redhat.com/rhel/7.3 |
+| **rhel7.4** | Red Hat Enterprise Linux 7.4 | 7.4 | linux | http://redhat.com/rhel/7.4 |
+| **win2k12r2** | Microsoft Windows Server 2012 R2 | 6.3 | winnt | http://microsoft.com/win/2k12r2 |
+| **win2k16** | Microsoft Windows Server 2016 | 1709 | winnt | http://microsoft.com/win/2k16 |
+
+
+### Use with presets
+
+A Virtual Machine Preset representing an operating system with a ``kubevirt.io/os-short-id`` label could be applied on any given
+Virtual Machine that have and match the``kubevirt.io/os-short-id`` label.
```
masayag: pls squash this line into the previous

masayag: Could you add a link from this section to the VM preset guide (if exists) ?

yanirq: >Could you add a link from this section to the VM preset guide (if exists) ?

once the presets user guide will be merged

guest-os-info.md
```diff
@@ -0,0 +1,34 @@
+## Guest Operating System Information
+
+Guest operating system identity for the Virtual Machine will be provided by the label ``kubevirt.io/os-short-id`` :
```
pkliczewski: As far as I understand we are going to use `os-short-id` not only for VMs. Please provide information on other places as well.

guest-os-info.md
```diff
@@ -0,0 +1,34 @@
...
+
+```
+metadata:
+  name: myvm
+  labels:
+    kubevirt.io/os-short-id: win2k12r2
+```
+
+The ``kubevirt.io/os-short-id`` label is based on the short OS identifier from [libosinfo](https://libosinfo.org/)
```
rmohr: What is the rationale for staying with that clunky name instaed of `kubevirt.io/os`?

yanirq: On one hand it is a clunky name, on the other hand it keeps a logical connection to the origin of the values the label can have  - an OS presentation in short id . Using os-short-id might also prevent confusion with general os type (linux/windows) for example.

guest-os-info.md
yanirq: Right, so we might want to move this to a more general section and provide links from Virtual Machine section in the user guide and in the future also from offline virtual machines and templates if needed.
@fabiand  any thoughts ?

rmohr: For sure you can apply them wherever you like, but as far as I understand this, this is about the use in combination with vm-presets, and presets only care about vms (if you apply it on a template it has no meaning for a preset).

guest-os-info.md
rmohr: I would then argue that I would either like to see one of the following:

* `kubevirt.io/os` may contain things like `win2k12r2`, `fedora27` or a family like `windows`.
* `kubevirt.io/os` and `kubevirt.io/os-family` to keep them separate and short, however why would you apply both or contradicting values? If I know it is fedora, I can apply fedora. If I just know it is `linux`, I just apply `linux`.

Especially since we don't plan to introduce `long-id`, it is strange to have it. If you mean by `origin` the connection to libosinfo, that is not really interesting for the end-user.

yanirq: ok, so lets have the ``kubevirt.io/os`` label instead of ``kubevirt.io/os-short-id``. 
for non end-users it would be helpful to know what it is based on - but for that i have added a developers doc. 
and if we will require extra labels such as os-family then we will add them later on. 
(main idea with short-id is not to overload labeles with os info and use it to query libosinfo database for additional info - but as you mentioned, this is not for end users)

guest-os-info.md
yanirq: @rmohr  we also thought about adding the label to offline VMs (with a combination of presets if needed) , but this was not finalized yet.
This could be changed in the future if we do so, and keep the current PR as is for now.

guest-os-info.md
pkliczewski: BTW, at the moment we use `annotations` to get `os-family`. From client perspective there is value in having it.

rmohr: @pkliczewski so this is for visualisation and orientation? Sounds good.

pkliczewski: @rmohr correct

yanirq:
@rmohr  do you still require more changes here ?

guest-os-info.md
```diff
@@ -0,0 +1,34 @@
...
+| Short ID | Name | Version | Family | ID |
+| --- | --- | --- | --- | --- |
+| **fedora26** | Fedora 26 | 26 | linux | http://fedoraproject.org/fedora/26 |
+| **fedora27** | Fedora 27 | 27 | linux | http://fedoraproject.org/fedora/27 |
+| **rhel7.0** | Red Hat Enterprise Linux 7.0 | 7.0 | linux | http://redhat.com/rhel/7.0 |
+| **rhel7.1** | Red Hat Enterprise Linux 7.1 | 7.1 | linux | http://redhat.com/rhel/7.1 |
+| **rhel7.2** | Red Hat Enterprise Linux 7.2 | 7.2 | linux | http://redhat.com/rhel/7.2 |
+| **rhel7.3** | Red Hat Enterprise Linux 7.3 | 7.3 | linux | http://redhat.com/rhel/7.3 |
+| **rhel7.4** | Red Hat Enterprise Linux 7.4 | 7.4 | linux | http://redhat.com/rhel/7.4 |
+| **win2k12r2** | Microsoft Windows Server 2012 R2 | 6.3 | winnt | http://microsoft.com/win/2k12r2 |
```
rmohr: As far as I can see, win2k12r2 is the only one we currently have a PR for.

yanirq: >As far as I can see, win2k12r2 is the only one we currently have a PR for.

we have this table under developer documents. so i can narrow this doc only for the existing PR and maintain it as we go along with new PRs.

rmohr: Yes please. :+1: 
As a general rule, we only document what we have, not what we plan to do.

yanirq:
@rmohr  document now includes only supported os labels  + use of os labels with presets example

fabiand:
lgtm

And the core examples have been merged.

PR #28, opened 2018-03-01
Add anchors to user-guide pages

rmohr:
On longer documentation pages it is hard to point people to the right section. Let's add the anchors plugin, so that we can share links to headings inside a page.

`gitbook install` will install the plugin. On gitbook.com this happens implicitly.


fabiand:
gtlm

PR #41, opened 2018-03-23
Update VM Preset user guide to reflect conflict logic

stu-gott:


vm-presets.md
```diff
@@ -140,6 +141,16 @@ metadata:
...
 ```
 
+Conflicts
+------------------------
+
+When multiple `VirtualMachinePresets` match a particular `VirtualMachine`, if
+they specify the same settings within a Domain, those settings must match. If
+two `VirtualMachinePresets` have conflicting settings (e.g. for the number of
+CPU cores requested), an error will occur, and the `VirtualMachine` will enter
+the `Failed` state, and a `Warning` event will be emitted explaining which
```
rmohr: The  going to failed behaviour confused me. Also have a look at: https://github.com/kubevirt/kubevirt/issues/805#issuecomment-378205710. But the description seems to be accurate.

stu-gott: @rmohr Is this still a concern to you?

PR #45, opened 2018-04-04
Add launching OfflineVM via `oc` command

gszasz:
This PR adds instructions how to call `oc` to create and launch an `OfflineVirtualMachine` from an
OpenShift template.  The workflow is covered by following pages:

- OfflineVM user guide 
- Basic Life-cycle
- OpenShift integration user guide

The latter was originally suggested by @stu-gott . It is a blueprint for a standalone documentation page covering all OpenShift-related KubeVirt features.


rmohr:
I would stick with kubectl for as long as there is no difference in the usage between kubectl and oc and only point out differences where they occure.

rmohr:
> @rmohr The card @gszasz is working on, is to add oc specific tests.
> I do think, that where kubectl & oc are the same, we can do table tests and support running both (will need to make sure the oc ones are only executed on the openshift env), but oc is important from QE perspective, so we cannot only test kubectl

I think that there is a misunderstanding. I am talking about the user guide. I don't think it makes much sense to bloat the guide with identical commands for `kubectl` and `oc`. For as long as they match we can stick with `kubectl`. That is what I meant.

nellyc:
@rmohr, cool, thanks for the clarification

openshift-integration.md
```diff
@@ -0,0 +1,105 @@
+# The OpenShift integration user guide
```
rmohr: Can you add that page to the `summary.md` page, so that it is added to the menu?

workloads/templates/README.md
```diff
@@ -0,0 +1,6 @@
+# Templates
+
+!> This only works on OpenShift so far \(See [Installation Guide](installation.md) for more information on how to deploy KubeVirt on OpenShift\).
+
+
+By deploying KubeVirt on top of OpenShift can user benefit from the [OpenShift template](https://docs.openshift.org/latest/dev_guide/templates.html) functionality.
```
fabiand: OpenShift **T**emplate

workloads/templates/vm-creation.md
```diff
@@ -0,0 +1,106 @@
...
+!> Only `OfflineVirtualMachine` object is currently supported.
+
+
+## Example template
+
+Here is an example template that defines an instance of the `OfflineVirtualMachine` object:
+
+\([cluster/vm-template-fedora.yaml](cluster/vm-template-fedora.yaml)\):
+
+```yaml
```
fabiand: For readability, shorten this example to the bare minimal informations required to illustrate the teplate (it does not need to work by copy and pasting). Just keep the link above to a working example.

Things to keep:
- fields where the parameters are used
- one or two fileds where NO parameters are used (to show that we can have non-parametrized fields)

workloads/templates/vm-creation.md
```diff
@@ -0,0 +1,106 @@
+# Virtual Machine Creation
+
+## Overview
+
+KubeVirt adds new types to the Kubernetes API to manage Virtual Machines and, in OpenShift, you can interact with these new resources via `oc` command as you would with any other Kubernetes API resource.
```
fabiand: Stroke this line - This is pretty redundant to other parts of the docs.

Please instead add, a not to long sentence along the lines of:

> OpenShift's template mechanism allows a user to create a set of objects from a template. KubeVirt is using this as a template mechanism for new OfflineVirtualMachines. The following chapter concentrates on the specific use around OfflineVirtualMachines, more informations on the templating mechanism can be [found in the OpenShift guide](https://docs.openshift.org/latest/dev_guide/templates.html).

Important points:
- Tell where Templates come from and what they do - link to the details
- Tell what we are using template for

Then (as you do) continue below with a detailed description.

workloads/templates/vm-creation.md
```diff
@@ -0,0 +1,106 @@
...
+              - name: disk1
+                volumeName: cloudinitvolume
+                disk:
+                  bus: virtio
+        volumes:
+          - name: registryvolume
+            registryDisk:
+              image: kubevirt/fedora-cloud-registry-disk-demo:latest
+          - name: cloudinitvolume
+            cloudInitNoCloud:
```
fabiand: Definetly drop this long line :)

workloads/templates/vm-creation.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+The JSON file is usually applied directly by piping the processed output to `oc create` command.
+
+```bash
+$ oc process -f cluster/vm-template-fedora.yaml \
+    -p NAME=testvm \
+    -p CPU_CORES=2 \
+    | oc create -f -
+```
```
fabiand: Please insert the output line of that command here

i.e.

```
$ … 
Objects foobar were created
```

Not sure what oc apply will output.

workloads/templates/vm-creation.md
```diff
@@ -0,0 +1,106 @@
...
+
+The command above results in creating a Kubernetes object according to the specification given by the template \(in this example it is an instance of the OfflineVirtualMachine object\).
+
+
+## Starting virtual machine from the created object
+
+The virtual machine can be then started by patching Kubernetes API resources.
```
fabiand: You can keep this example.
But you should also:
- point out that it's now a regular OVM
- reference the life-cycle page and tell the user to continue reading there

PR #48, opened 2018-04-07
Minor edits

rwsu:
A few minor corrections to spelling and grammar. 

Improve readability by adding abbreviations and making them capitalized.

### DIFF
--- a/workloads/controllers/virtual-machine-replica-set.md
+++ b/workloads/controllers/virtual-machine-replica-set.md
@@ -1,36 +1,36 @@
-# Virtual Machine Replica Set
+# VirtualMachineReplicaSet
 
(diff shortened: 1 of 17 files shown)


workloads/controllers/virtual-machine-replica-set.md
```diff
@@ -1,36 +1,36 @@
-# Virtual Machine Replica Set
+# VirtualMachineReplicaSet
 
-\#\# Virtual Machine Replica Set
+## VirtualMachineReplicaSet
```
rmohr: We should comletely remove that line

rmohr: We should comletely remove that line

PR #61, opened 2018-05-08
Update service integration documentation

rmohr:
Focus more on how to expose the VM, than on how to expose the underlying pod. Tries to address confusions on how well we can integrate with core-services, although we are using CRDs and not an alternative container runtime for virtualization workloads.

### DIFF
--- a/workloads/controllers/README.md
+++ b/workloads/controllers/README.md
@@ -1,4 +1,7 @@
 # Controllers
 
-Controllers provide the logic to manage virtual machine instances in a way that addresses specific use-cases.
+Controllers provide the logic to manage virtual machine instances in a way that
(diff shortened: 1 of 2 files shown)


duyanyan:
@rmohr I just have a minor question that could we keep at least one example of create service by exposing the underlying pod, maybe we could put it in Notes or somewhere. Because it is really a more convenient way to create service than using a service yaml/json file. WDYT? Others lgtm

rmohr:
Yes I can keep one. I was hoping that https://github.com/kubevirt/kubevirt/pull/962 is ready soon, so that we have a  convenient replacement.

rmohr:
@duyanyan done.

PR #66, opened 2018-05-14
Use convenience command for creating the right k8s secret for cloud-init

rmohr:
@davidvossel directly seeing the secret is also valuable, but it may be easier for people to use the secrets if we show them the high-level k8s command. What do you think?

### DIFF
--- a/workloads/virtual-machines/startup-scripts.md
+++ b/workloads/virtual-machines/startup-scripts.md
@@ -143,16 +143,7 @@ echo "Hi from startup script!"
 END
 
 # Store the startup script in a Kubernetes Secret
-
-cat << END > my-secret.yaml
-apiVersion: v1
-kind: Secret
-metadata:
-  name: my-vm-secret
-type: Opaque
-data:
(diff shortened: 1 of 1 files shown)


stu-gott:
In this case, @rmohr's approach is the way to go.

@davidvossel would it be helpful to have a separate document to briefly discuss secrets in general?

For the sake of enlightening users on how to create secrets, the `--from-literal` flag is also very useful.

`kubectl create secret generic example-secret --from-literal=foo=base64encoded_string`

would yield

```yaml
apiVersion: v1
data:
  foo: YmFzZTY0ZW5jb2RlZF9zdHJpbmc=
kind: Secret
metadata:
  name: example-secret
  namespace: default
...
type: Opaque
```

But trying to add all that to the narrative of this document can dilute the real point of it.

rmohr:
> Maybe show the result of the secret post?

I would agree with @stu-gott that it would be a better story if we show the convenience method only. I think it is  simply a more convincing story on how easy it actually it is to use secrets for that.

PR #82, opened 2018-06-11
Add details how to configure hugepages on a VM

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/workloads/virtual-machines/virtualized-hardware-configuration.md
+++ b/workloads/virtual-machines/virtualized-hardware-configuration.md
@@ -278,3 +278,38 @@ Various VM resources, such as a video adapter, IOThreads, and supplementary syst
 
 Please see [how Pods with resource requests are scheduled](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-requests-are-scheduled) for additional information on resource requests and limits.
 
+## Hugepages
+
+KubeVirt give you possibility to use hugepages as backing memory for your VM. You will need to provide desired amount of memory `resources.requests.memory` and size of hugepages to use `memory.hugepages.pageSize`, for example for x86_64 architecture it can be `2Mi`.
+
+```yaml
+apiVersion: kubevirt.io/v1alpha1
+kind: VirtualMachine
+metadata:
+  name: myvm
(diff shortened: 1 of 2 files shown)


workloads/virtual-machines/virtualized-hardware-configuration.md
```diff
@@ -278,3 +278,37 @@ Various VM resources, such as a video adapter, IOThreads, and supplementary syst
...
+apiVersion: kubevirt.io/v1alpha1
+kind: VirtualMachine
+metadata:
+  name: myvm
+spec:
+  domain:
+    resources:
+      requests:
+        memory: "64Mi"
+    hugepages:
```
fabiand: Instead of a top-level hugepages item it looks more sane to have it as a parameter to a `memory` struct, in order to give it some smenatics, iow:

```yaml
spec:
  domain:
    resources:
      requests:
        memory: "64Mi"
    memory:
      hugepages:
        size: "2Mi"
```

cynepco3hahue: @fabiand I thought about it, but it looks a little bit confusing to have `resources.requests.memory` and `memory.hugepages`
What do you think?

fabiand: We need to look at the VM.

For the VM, regardless if it's plain memory or hugepage, the VM will always just get memory.
That is why it's okay that we only have a memory request field. Because it's about the vm memory, not about what we technically need on the pod to implement this.

workloads/virtual-machines/virtualized-hardware-configuration.md
```diff
@@ -278,3 +278,38 @@ Various VM resources, such as a video adapter, IOThreads, and supplementary syst
...
+metadata:
+  name: myvm
+spec:
+  domain:
+    resources:
+      requests:
+        memory: "64Mi"
+    memory:
+      hugepages:
+        pageSize: "2Mi"
```
rmohr: Can you explicitely call out the mentioned fields in the verbal description? `resources.requests.memory` and memory.hugepages.pageSize`?

PR #86, opened 2018-06-14
Add note about CRIO

fabiand:
Add a note about supported container runtimes.

Fixes: kubevirt/kubevirt#896

Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>

### DIFF
--- a/installation/README.md
+++ b/installation/README.md
@@ -10,12 +10,18 @@ A few requirements need to be met before you can begin:
 * `kubectl` client utility
 * `git`
 
-### Minimum Requirements
+### Minimum Cluster Requirements
 
-| Component | Minimum Version |
-| --- | --- |
-| Kubernetes | 1.7 |
-| KubeVirt | [v0.1.0](https://github.com/kubevirt/kubevirt/releases/v0.1.0) |
+Kubernetes 1.8 or later is required to run KubeVirt.
+
+In addition it can be that feature gates need to be opened.
+
+#### Runtime
+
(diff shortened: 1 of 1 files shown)


installation/README.md
```diff
@@ -10,12 +10,18 @@ A few requirements need to be met before you can begin:
...
 * `git`
 
-### Minimum Requirements
+### Minimum Cluster Requirements
 
-| Component | Minimum Version |
-| --- | --- |
-| Kubernetes | 1.7 |
-| KubeVirt | [v0.1.0](https://github.com/kubevirt/kubevirt/releases/v0.1.0) |
+Kubernetes 1.8 or later is required to run KubeVirt.
```
rmohr: Wouldn't 1.9 be more accurate?

installation/README.md
```diff
@@ -10,12 +10,18 @@ A few requirements need to be met before you can begin:
...
+
+In addition it can be that feature gates need to be opened.
```
rmohr: Hm, we need to be more demanding here. Especially the validation webhook and some certificate features just needs to be enabled. But better than before.

fabiand: Absolutely.

We actually have https://github.com/kubevirt/kubevirt/issues/1023 to do this properly. We should have the same list here.

I merely added the few words to at least give head's up that something might be needed in addition.

I could link the issue ...

installation/README.md
fabiand: True

PR #89, opened 2018-06-15
Document the new `model` interface attribute

booxter:
With https://github.com/kubevirt/kubevirt/pull/1167 merged, kubevirt will
support model attribute for interfaces (before the PR, it was supported via
annotations but we never captured that in documentation, so no accommodation
needed).

### DIFF
--- a/workloads/virtual-machines/interfaces-and-networks.md
+++ b/workloads/virtual-machines/interfaces-and-networks.md
@@ -60,7 +60,13 @@ This backend type is used if an interface should be connected to the regular pod
 
 In some cases the underlying network plugin (flannel, weave, OpenShift SDN) acts as an Ethernet bridge or switch, in those cases the `pod` backend can also be used to provide an IP level connectivity to an interface (see backend `bridge.delegateIP`).
 
-## Available connections methods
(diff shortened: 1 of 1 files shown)


booxter:
@fabiand @vladikr  please review (of course, after the kubevirt PR is ready to merge).

booxter:
@vladikr we've landed the kubevirt part, so it would be nice to have it documented. Please review.

rmohr:
The information seems to fit into the schema of this document. Can we follow a little bit more the structure like we have for https://kubevirt.io/user-guide/#/workloads/virtual-machines/disks-and-volumes? The structure there is lent from k8s, which I personally find pretty easy to follow and user-focused.

So to have 
 * Examples per configs
 * For instance also show where and how to set the model
 * For a comprehensive and up-to-date list of supported values, link to the api-reference and make sure that the api-reference comment contains supported values.

It may just be me, but I find the structure of this document very confusing and it is almost impossible for me to parse it and extract the necessary information. It also looks a little bit like this documentation tries to be a mixture of api-reference and user-guide. We should leverage the fact that we have an api-reference.

Issue #91, opened 2018-06-21
At least to me the networking documentation is hard to follow


rmohr:
>Can we follow a little bit more the structure like we have for https://kubevirt.io/user-guide/#/workloads/virtual-machines/disks-and-volumes? The structure there is lent from k8s, which I personally find pretty easy to follow and user-focused.
>
>So to have 
> * Examples per configs
> * For instance also show where and how to set the model
> * For a comprehensive and up-to-date list of supported values, link to the api-reference and make sure that the api-reference comment contains supported values.
>
> It may just be me, but I find the structure of this document very confusing and it is almost impossible for me to parse it and extract the necessary information. It also looks a little bit like this documentation tries to be a mixture of api-reference and user-guide. We should leverage the fact that we have an api-reference.

from https://github.com/kubevirt/user-guide/pull/89#issuecomment-399029970

fabiand:
I totally agree. The networking do is structural wise not so good.

Technically the problem s that we have three components: frontend, connection, backend.

This makes the current situation so bad.

But we can probably flatten this by mentioning the relevant connection times in the context of frontend, then we again only have frontends and backends, which should allow us to follow the disk pattern.

My 2ct

booxter:
@rmohr Agreed. About examples, how do we guarantee they are valid / still working as we move forward? Maybe we could consume them from kubevirt repo (cluster/examples/?) and allow validations to happen against each of examples. What do you think?

fabiand:
> Maybe we could consume them from kubevirt repo (cluster/examples/?) and allow validations to happen against each of examples. What do you think?

Technically this works with docsify.

An exampe is here:
https://github.com/kubevirt/user-guide/pull/92

Gotot th evirtual machines section to see bar and an inline external yaml file.

rmohr:
> @rmohr Agreed. About examples, how do we guarantee they are valid / still working as we move forward? Maybe we could consume them from kubevirt repo (cluster/examples/?) and allow validations to happen against each of examples. What do you think?

Anything that helps to keep them valid is great. I did not have a clear picture in my mind on how to do that. If you have ideas go ahead! :+1:

booxter:
Please assign me to the issue (or give me access to the action). I will look into this during / after macAddress work.

PR #92, opened 2018-06-21
[POC] Inline external

fabiand:
This is an example how an external file can be inlined:

Can be previewed here:
https://fabiand.github.io/user-guide/#/workloads/virtual-machines/README

Described here:
https://docsify.js.org/#/embed-files

Relates to https://github.com/kubevirt/user-guide/issues/91#issuecomment-399109809

### DIFF
--- a/workloads/virtual-machines/README.md
+++ b/workloads/virtual-machines/README.md
@@ -6,3 +6,5 @@ The `VirtualMachineInstance` type conceptionally has two parts:
 * Information about the virtual machine ABI
 
 Every `VirtualMachineInstance` object represents a single running virtual machine instance.
+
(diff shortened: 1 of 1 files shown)


fabiand:
@booxter fyi

booxter:
I am now not sure inlining would be helpful. It would mean that we need to inline complete examples into docs. The examples may be quite long and loose the right focus on specific field described in a particular place of the document.

fabiand:
That's true - good perspective on the matter.

Well - maybe they can extend each other: Shortened snippets (extensive use of ellipsis) and links to working exampples (which live in the kubevirt repo).

But this means that this issue is not needed.

PR #102, opened 2018-07-06
Document annotations and labels of user interest

booxter:
Things to focus when reviewing:

- whether the index.html attachment makes sense (I couldn't find a better place to attach the info; and I didn't feel like creating more levels of embedding);
- an argument can be made that this information should be as close to API reference docs as possible (perhaps part of them).

Related to kubevirt/kubevirt#1080


booxter:
@rmohr This is user-guide part of annotations / labels docs.

PR #103, opened 2018-07-11
Escape colons in MAC addresses

booxter:
Markdown format has special meaning for :term:s and trashes MAC
addresses in HTML output unless we escape.


booxter:
The issue can be seen in: https://kubevirt.io/user-guide/#/workloads/virtual-machines/interfaces-and-networks?id=frontend

PR #106, opened 2018-07-18
Fix cpu-model documentation

rmohr:
The field for setting the cpu model is called `model` and not `mode`.

Fixes https://github.com/kubevirt/kubevirt/issues/1340.


rmohr:
Possibly wrong fix. Need to check again.

fabiand:
Yes, it's two things - mode and model.

PR #107, opened 2018-07-18
Fix cpu-model documentation

rmohr:
We don't support cpu-mode right now. Only `model`. If not specified, KubeVirt falls back to host-model.

Aligns the documentation with the code. Related to https://github.com/kubevirt/kubevirt/issues/1340

### DIFF
--- a/workloads/virtual-machines/virtualized-hardware-configuration.md
+++ b/workloads/virtual-machines/virtualized-hardware-configuration.md
@@ -112,24 +112,10 @@ spec:
 ```
 
 You can check list of available models [here](https://github.com/libvirt/libvirt/blob/master/src/cpu/cpu_map.xml).
- 
-### Mode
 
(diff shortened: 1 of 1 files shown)


rmohr:
Ok, so this may not solve https://github.com/kubevirt/kubevirt/issues/1340 but it at least describes what we have. So we don't have `host-passthrough` right now. Only `host-model`, if no `model` was specified.

rmohr:
@fabiand can it be that you by accident pushed the `mode` cpu documentation to the user-guide instead of your fork: https://github.com/kubevirt/user-guide/commit/46104b476856fc67ab9c5088d4ea86b54739ea4e?

fabiand:
Which means that host-passthrough can not be set ATM.

cynepco3hahue:
@fabiand Sorry for miss-understanding I thought we talked only about CPU models and not about host-passthrough, will close gap ASAP.

PR #124, opened 2018-09-25
Documentation for IOThreads feature

stu-gott:
Documentation for https://github.com/kubevirt/kubevirt/pull/1480


stu-gott:
Good catch @petrkotas! I'll fix that

PR #129, opened 2018-10-12
Add data corruption documentation to VMIRS

rmohr:
Make clear that VMIRS can corrupt disks if you are not cautious.

### DIFF
--- a/workloads/controllers/virtual-machine-replica-set.md
+++ b/workloads/controllers/virtual-machine-replica-set.md
@@ -28,21 +28,56 @@ All service exposure options that apply to a VirtualMachineInstance apply to a V
 
 ## When to use a VirtualMachineInstanceReplicaSet
 
-Using VirtualMachineInstanceReplicaSet is the right choice when one wants many identical VMs and does not care about maintaining any disk state after the VMs are terminated.
+> **Note:** The base assumption is that referenced disks are read-only or that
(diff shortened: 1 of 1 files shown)


workloads/controllers/virtual-machine-replica-set.md
```diff
@@ -28,7 +28,25 @@ All service exposure options that apply to a VirtualMachineInstance apply to a V
 
 ## When to use a VirtualMachineInstanceReplicaSet
 
-Using VirtualMachineInstanceReplicaSet is the right choice when one wants many identical VMs and does not care about maintaining any disk state after the VMs are terminated.
+> **Note:** The base assumption is that referenced disks are read-only or that
+> the VMIs are writing internally to a tmpfs. The most obvious volume sources
+> for VirtualMachineInstanceReplicaSets which KubeVirt supports are referenced
+> below. If other types are used **datacorruption** is possible.
```
stu-gott: data corruption is probably two words?

rmohr: @stu-gott thanks. Fixed.

rmohr:
> should we go as far as refusing to accept VMIRS's with unsafe storage in the validation webhook?

Maybe. For as long as the VMIs tread the disks as read-only or write to a tmpfs it would however work. I am uncertain.

PR #131, opened 2018-10-18
hostDisk doc

mfranczy:


mfranczy:
@rmohr could you take a look and check if it makes sense?
@stu-gott could you check from native speaker perspective?

mfranczy:
I forgot about PVC initialization, added to PR.

workloads/virtual-machines/disks-and-volumes.md
```diff
@@ -376,6 +379,47 @@ spec:
...
 
+### hostDisk
+
+A ```hostDisk``` volume type provides the ability to create or use a disk image located somewhere on a node.
+It works similar to a ```hostPath``` in Kubernetes and provides two usage types:
+
+* ```DiskOrCreate``` if a disk image does not exist at a given location then create one
+* ```Disk``` a disk image must exist at a given location
+
+Example: Create a disk image with 1Gi size at /data/disk.img location and attach it to a VM.
```
stu-gott: Suggest rewording to:

Example: Create a 1Gi disk image located at /data/disk.img and attach it to a VM.

mfranczy: done

PR #132, opened 2018-10-19
Added block pvc description

slintes:


slintes:
For the VMI manifest there is no difference. Not sure about adding PV/PVC manifests, as they depend on your storage driver. And the example from the demo yesterday only works on our test cluster, which provides the cirros device...?

stu-gott:
Oops my bad. In my mind there were changes visible at the VMI layer. you are correct!

PR #137, opened 2018-10-31
disks cache doc

mfranczy:


mfranczy:
@vladikr @fabiand can you take a look?

PR #142, opened 2018-11-06
document IOThreads pinning to pCPUs

vladikr:
Document how IOThreads are getting pinned to pCPUs on VMIs with dedicated CPUs
Related to https://github.com/kubevirt/kubevirt/pull/1604


fabiand:
Thanks!

PR #151, opened 2018-11-15
Document what users can do with the "common templates" subproject

fromanirh:
This is the initial batch of documentation. More is still to come (also from @sHaggYcaT ).
I'm submitting this PR now because, albeit not complete, the initial changes are self-sufficient and allow us to bootstrap the "common-templates.md" file, which will receive more content later.

CC @sHaggYcaT @MarSik for reviews/comments

### DIFF
--- a/workloads/templates/common-templates.md
+++ b/workloads/templates/common-templates.md
@@ -31,14 +31,16 @@ $ oc create -f https://github.com/kubevirt/common-templates/releases/download/$V
 ## Editable fields
 
 You can edit the fields of the templates which define the amount of resources which the VMs will receive.
-The list of fields includes:
 
-- spec.template.spec.domain.cpu.cores
-- spec.template.spec.domain.resources.requests.memory
-- spec.template.spec.domain.devices.disks
-- spec.template.spec.volumes
-- spec.template.spec.networks
+Each template can list a different set of fields that are to be considered editable.
+The fields are used as hints for the user interface, and also for other components in the cluster.
 
+The editable fields are taken from annotations in the template:
+```yaml
+metadata:
+  annotations:
+    template.cnv.io/editable: ...
+```
 
 ## Relationship between templates and VMs
 
--- a/workloads/templates/common-templates.md
+++ b/workloads/templates/common-templates.md
@@ -20,7 +20,7 @@ More documentation is available in the [common templates subproject](https://git
 ## Accessing the virtual machine templates
 
 If you installed KubeVirt using a supported method you should find the common templates preinstalled in the cluster.
-Should you want to upgrade the templates, or installed them from scratch, you can use one of the [supported release](https://github.com/kubevirt/common-templates/releases)
+Should you want to upgrade the templates, or install them from scratch, you can use one of the [supported releases](https://github.com/kubevirt/common-templates/releases)
 
 To install the templates:
 ```bash
--- a/workloads/templates/common-templates.md
+++ b/workloads/templates/common-templates.md
@@ -35,11 +35,33 @@ You can edit the fields of the templates which define the amount of resources wh
 Each template can list a different set of fields that are to be considered editable.
 The fields are used as hints for the user interface, and also for other components in the cluster.
 
-The editable fields are taken from annotations in the template:
+The editable fields are taken from annotations in the template. Here is a snippet presenting a couple of most
+commonly found editable fields:
+
 ```yaml
 metadata:
   annotations:
-    template.cnv.io/editable: ...
+    template.cnv.io/editable: |
+      /objects[0].spec.template.spec.domain.cpu.cores
+      /objects[0].spec.template.spec.domain.resources.requests.memory
+```
+
+Each entry in the editable field list must be a [jsonpath](https://kubernetes.io/docs/reference/kubectl/jsonpath/).
+The actually editable field is the last entry (the "leaf") of the path. For example, the following minimal snippet highlights
+the fields which you can edit:
+```yaml
+objects:
+  spec:
+    template:
+      spec:
+        domain:
+          cpu:
+            cores:
+              VALUE # this is editable
+          resources:
+            requests:
+              memory:
+                VALUE # this is editable
 ```
 
 ## Relationship between templates and VMs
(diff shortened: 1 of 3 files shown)


workloads/templates/common-templates.md
```diff
@@ -0,0 +1,38 @@
...
+## What is a virtual machine template?
+
+The KubeVirt projects provides a set of [templates](https://docs.okd.io/latest/dev_guide/templates.html) to create VMs to handle common usage scenarios.
+These templates provide a combination of some key factors that could be further customized and processed to have a Virtual Machine object.
+The key factors which define a template are
+- Workload
+most Virtual Machine should be *generic* to have maximum flexibility; the *highperformance* workload trades some of this flexibility to
+provide better performance
+- Guest Operating System (OS)
+Enable optimizations depending on the OS which is going to run on the guest.
```
MarSik: Also makes sure the default emulated hardware is compatible with the OS.

sHaggYcaT: Thanks @fromanirh, I've add reference to your doc from my own (I've just add it to the PR)

sHaggYcaT: @fromanirh how do you think, maybe good idea to add information regarding webui? Just something very short - that it's exist and support templates?
You can check my paragraph: https://github.com/sHaggYcaT/user-guide/blob/8epic-templating-14-user-guide-section-for-the-use-of-templates-initiative-document-happy-path/workloads/templates/vm-creation.md
I not sure is it information enough, or otherwise, more then enough.

fromanirh: @sHaggYcaT I think what you added is just fine. It mentions the existing web-ui, so users can look up for it if they want, in the web ui official docs.

fromanirh: > Also makes sure the default emulated hardware is compatible with the OS.

done (and rephrased)

workloads/templates/common-templates.md
```diff
@@ -0,0 +1,55 @@
...
+  of the VM, and allows performance optimizations.
+- Size (flavor) 
+  Defines the amount of resources (CPU, memory) to allocate to the VM.
+
+More documentation is available in the [common templates subproject](https://github.com/kubevirt/common-templates)
+
+## Accessing the virtual machine templates
+
+If you installed KubeVirt using a supported method you should find the common templates preinstalled in the cluster.
+Should you want to upgrade the templates, or installed them from scratch, you can use one of the [supported release](https://github.com/kubevirt/common-templates/releases)
```
MarSik:
```suggestion
Should you want to upgrade the templates, or install them from scratch, you can use one of the [supported releases](https://github.com/kubevirt/common-templates/releases)
```

workloads/templates/common-templates.md
```diff
@@ -0,0 +1,55 @@
...
+
+To install the templates:
+```bash
+$ export VERSION="v0.3.1"
+$ oc create -f https://github.com/kubevirt/common-templates/releases/download/$VERSION/common-templates-$VERSION.yaml
+```
+
+## Editable fields
+
+You can edit the fields of the templates which define the amount of resources which the VMs will receive.
```
MarSik: Each template can list a different list of fields that are to be considered editable. It serves as a hint to the UI and to whatever upgrade logic we will have. So either frame the list below as an example (and make it more obvious it is just an example) or remove it please. Also you should actually say where the list is defined (the annotation name).

workloads/templates/common-templates.md
```diff
@@ -0,0 +1,55 @@
...
+- spec.template.spec.domain.cpu.cores
+- spec.template.spec.domain.resources.requests.memory
+- spec.template.spec.domain.devices.disks
+- spec.template.spec.volumes
+- spec.template.spec.networks
+
+
+## Relationship between templates and VMs
+
+Once [processed](), the templates produce VM objects to be used in the cluster. The VMs produced from templates will have a `vm.cnv.io/template` label, whose
```
MarSik: Missing link from processed? Also, can you show an example of how to actually create a VM based on the label?

yanirq: I agree, we should add an example of vm creation after we either created all/some/one of the templates with the proper parameters, e.g. for the template rhel7-generic-tiny :  `oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -`

sHaggYcaT: Is it what you looking for: https://github.com/fromanirh/user-guide/pull/1
?

fromanirh: Not sure what you mean with "create a VM based on the label", please check my last commit and let's move forward from there.

MarSik: That was a typo :) I wanted to say based on the template.

workloads/templates/common-templates.md
MarSik: This is still unresolved.

fromanirh: yes, I missed this comment. On it now.

workloads/templates/common-templates.md
```diff
@@ -31,14 +31,16 @@ $ oc create -f https://github.com/kubevirt/common-templates/releases/download/$V
...
-- spec.template.spec.networks
+Each template can list a different set of fields that are to be considered editable.
+The fields are used as hints for the user interface, and also for other components in the cluster.
 
+The editable fields are taken from annotations in the template:
+```yaml
+metadata:
+  annotations:
+    template.cnv.io/editable: ...
+```
```
MarSik: Please leave an example path there (cpu and memory perhaps) and explain what the one per line paths mean (described here: https://github.com/kubevirt/common-templates/blob/master/templates/README.md#templateyaml). Jsonpath should be a link and point to https://kubernetes.io/docs/reference/kubectl/jsonpath/ I think.

> The content is a line separated list of jsonpath selectors.
> The jsonpath root is the objects: element of the template

fromanirh: done

rmohr:
@fromanirh will merge it once you add it to e.g. https://github.com/kubevirt/user-guide/blob/master/_sidebar.md.

PR #165, opened 2018-12-04
Template docs custom

fromanirh:
This PR adds documentation about customization of the common templates. Contributions by @sHaggYcaT , I (@fromanirh ) am just handling the PR process.

### DIFF
--- a/workloads/templates/common-templates.md
+++ b/workloads/templates/common-templates.md
@@ -47,6 +47,7 @@ metadata:
 ```
 
 Each entry in the editable field list must be a [jsonpath](https://kubernetes.io/docs/reference/kubectl/jsonpath/).
+The jsonpath root is the objects: element of the template.
 The actually editable field is the last entry (the "leaf") of the path. For example, the following minimal snippet highlights
 the fields which you can edit:
 ```yaml
@@ -134,43 +135,87 @@ in templates do not automatically affect VMs, or vice versa).
 
 ## common template customization
 
-### template customization - memory, CPU
+The templates provided by the kubevirt project provide a set of conventions and annotations that augment the basic feature of the
+[openshift templates](https://docs.okd.io/latest/dev_guide/templates.html).
+You can customize your kubevirt-provided templates editing these annotations, or you can add them to your existing templates to
+make them consumable by the kubevirt services.
 
-There are three options to customize VM memory and CPU:
+Here's a description of the kubevirt annotations. Unless otherwise specified, the following keys are meant to be top-level entries
+of the template metadata, like
 
- * select flavor - tiny, small, medium, etc. Each flavor grants to VM different amount of RAM and CPU cores.
- * setting up directly editable fields spec.template.spec.domain.cpu.cores and spec.template.spec.domain.resources.requests.memory, example:
-
-```bash
-oc patch virtualmachine testguest --type merge -p '{"spec":{"template":{"spec":{"domain":{"cpu":{"cores":'3'}}}}}}'
-virtualmachine.kubevirt.io/testguest patched
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: windows-10
+  annotations:
+    openshift.io/display-name: "Generic demo template"
 ```
- * use WebUI: WebUI supports different flavors and Workload Profiles. Also it supports custom amount of memory and CPUs
 
-### template customization - networking
+All the following annotations are prefixed with `defaults.template.cnv.io`, which is omitted below for brevity. So the actual annotations you should use will look like
 
-There are two options to customize Networking in VM:
- * setting up directly editable field(see example above): spec.template.spec.networks. You can edit following:
-   * name (of interface)
-   * pod
- * edit networking in webUI. Please note: WebUI also edits spec.domain.interfaces and supports changing mac address
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: windows-10
+  annotations:
+    defaults.template.cnv.io/disk: default-disk
+    defaults.template.cnv.io/volume: default-volume
+    defaults.template.cnv.io/nic: default-nic
+    defaults.template.cnv.io/network: default-network
+```
 
-### template customization - disks
+Unless otherwise specified, all annotations are meant to be safe defaults, both for performance and compability, and hints for the CNV-aware UI and tooling.
 
-There are two options to customize disks:
+#### disk
+default disk type to use. It is recommended to aim for maximum compatibility.
 
- * Setting up sole variable - name of PersistentVolumeClaim(PVC)
- * Choose workload profile
- * Choose editable fields
+Example:
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: Linux
+  annotations:
+    defaults.template.cnv.io/disk: virtio
+```
 
-#### Setting up name of PVC
-Each template has only one variable for disk customization - name of PersistentVolumeClaim(PVC).  Note: PVC should exist before you start your VM at first time. 
+#### nic
+default nic model to use. Likewise the disk implementation, it is recomended to pick default which ensure the maximum compatibility.
 
-#### Choose worload profile
-In a case if high performance workload profile choosed, then kubevirt enables [IOThread](https://libvirt.org/formatdomain.html#elementsIOThreadsAllocation) qemu feature for disk. 
-Note: not each operation system has an option of high performance workload profile.  
+Example:
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: Windows
+  annotations:
+    defaults.template.cnv.io/nic: e1000
+```
 
-### Editable fields related with disks
-Each template has ability to direct edit editable fields (see details and example above). By fact it provides same ability to choose name of PVC in comparison with changing variable
+#### volume
+Default volume implementation to be used as backend for disks. The underlying volume implementation allow to configure the performance parameters (cache).
 
-Note: WebUI cant create specific disk in the Create Virtual Machine wizzard - only attach existing PVC
+Example:
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: Linux
+  annotations:
+    defaults.template.cnv.io/volume: custom-volume
+```
+
+#### network
+Default network to attach the NIC(s) to. Likewise the volume, the underlying network should be configured for the performance and the connectivity.
+
+Example:
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: Linux
+  annotations:
+    defaults.template.cnv.io/network: fast-net
+```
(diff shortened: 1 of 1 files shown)


workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
...
 in templates do not automatically affect VMs, or vice versa).
+
+## common template customization
+
+### template customization - memory, CPU
+
+There are three options to customize VM memory and CPU:
+
+ * select flavor - tiny, small, medium, etc. Each flavor grants to VM different amount of RAM and CPU cores.
+ * setting up directly editable fields spec.template.spec.domain.cpu.cores and spec.template.spec.domain.resources.requests.memory, example:
```
MarSik: Please make some of the references into links - the editable fields for example should point to the part that describes what it is and what is the format. Because you can customize the list of editable fields as well.

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
 
 Please note that, after the generation step, VM objects and template objects have no relationship with each other besides the aforementioned label (e.g. changes
 in templates do not automatically affect VMs, or vice versa).
+
+## common template customization
+
+### template customization - memory, CPU
+
+There are three options to customize VM memory and CPU:
```
MarSik: Your title says template customization, not VM customization. Selecting a flavor is not really relevant as that usually means a different template file.

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
...
+virtualmachine.kubevirt.io/testguest patched
+```
+ * use WebUI: WebUI supports different flavors and Workload Profiles. Also it supports custom amount of memory and CPUs
+
+### template customization - networking
+
+There are two options to customize Networking in VM:
+ * setting up directly editable field(see example above): spec.template.spec.networks. You can edit following:
+   * name (of interface)
+   * pod
```
MarSik: You can't edit them if the editable fields do not list them. Each template can have a different list of editable fields..
Also couple of examples would be nice.

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
...
+ * edit networking in webUI. Please note: WebUI also edits spec.domain.interfaces and supports changing mac address
+
+### template customization - disks
+
+There are two options to customize disks:
+
+ * Setting up sole variable - name of PersistentVolumeClaim(PVC)
```
MarSik: Again, you are describing customization of a VM and not the customization of the template here. There is no way to set variable in the template.

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
...
+ * Choose workload profile
+ * Choose editable fields
```
MarSik: How?

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
...
+   * name (of interface)
+   * pod
+ * edit networking in webUI. Please note: WebUI also edits spec.domain.interfaces and supports changing mac address
+
+### template customization - disks
+
+There are two options to customize disks:
+
+ * Setting up sole variable - name of PersistentVolumeClaim(PVC)
+ * Choose workload profile
```
MarSik: See above plus workload profiles have nothing to do with disk customization.

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
...
+ * Choose editable fields
+
+#### Setting up name of PVC
+Each template has only one variable for disk customization - name of PersistentVolumeClaim(PVC).  Note: PVC should exist before you start your VM at first time. 
+
+#### Choose worload profile
+In a case if high performance workload profile choosed, then kubevirt enables [IOThread](https://libvirt.org/formatdomain.html#elementsIOThreadsAllocation) qemu feature for disk.
```
MarSik: This is again an implementation detail that is related to the VM. But the section is called template customization.
Moreover IOthreads are just one possible optimization the templates enable.

MarSik: And anyway, you are not explaining the "how".

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
...
+
+#### Setting up name of PVC
+Each template has only one variable for disk customization - name of PersistentVolumeClaim(PVC).  Note: PVC should exist before you start your VM at first time. 
+
+#### Choose worload profile
+In a case if high performance workload profile choosed, then kubevirt enables [IOThread](https://libvirt.org/formatdomain.html#elementsIOThreadsAllocation) qemu feature for disk. 
+Note: not each operation system has an option of high performance workload profile.  
+
+### Editable fields related with disks
+Each template has ability to direct edit editable fields (see details and example above). By fact it provides same ability to choose name of PVC in comparison with changing variable
```
MarSik: example above. What example?

workloads/templates/common-templates.md
```diff
@@ -131,3 +131,46 @@ oc process rhel7-generic-tiny PVCNAME=mydisk NAME=rheltinyvm | oc apply -f -
 
 Please note that, after the generation step, VM objects and template objects have no relationship with each other besides the aforementioned label (e.g. changes
 in templates do not automatically affect VMs, or vice versa).
+
+## common template customization
```
MarSik: The section needs to describe what you can do to the template, which special annotations we have there, what labels are provided and what the meanings are. Creating a VM from the template can be in separate section if you wish, but we need to describe the format of the template.

fromanirh: OK, this means we should document most if not all[1] the annotations listed here: https://github.com/fromanirh/common-templates/blob/template-validation-docs/templates/README.md#templateyaml

this will duplicate some information between documents, but it looks like a necessary evil - AFAICT user doc shouldn't directly link developer doc, even thoug lines are a bit blurred for templates.

I'll also add a couple of "oc process patch" command examples only, being it strightforward to do for the other annotations.

+++

[1] for example the version field should be pretty self explanatory, maybe just a big red label "don't touch this"

workloads/templates/common-templates.md
fromanirh: I think we can remove the flavor references, and in general the content related to the VM customization.

workloads/templates/common-templates.md
fromanirh: Same decision as per flavor, VM customization-related content should go away (from this document)

fabiand:
Actually, it looks we still need updates from Francesco.

@fromanirh ping?

workloads/templates/common-templates.md
```diff
@@ -134,43 +135,87 @@ in templates do not automatically affect VMs, or vice versa).
...
-### template customization - networking
+Unless otherwise specified, all annotations are meant to be safe defaults, both for performance and compability, and hints for the CNV-aware UI and tooling.
 
-There are two options to customize Networking in VM:
- * setting up directly editable field(see example above): spec.template.spec.networks. You can edit following:
-   * name (of interface)
-   * pod
- * edit networking in webUI. Please note: WebUI also edits spec.domain.interfaces and supports changing mac address
+#### disk
+default disk type to use. It is recommended to aim for maximum compatibility.
```
MarSik: The value is not the disk type, but a name of disk that serves as a template.

workloads/templates/common-templates.md
```diff
@@ -134,43 +135,87 @@ in templates do not automatically affect VMs, or vice versa).
...
+kind: Template
+metadata:
+  name: Linux
+  annotations:
+    defaults.template.cnv.io/disk: virtio
+```
 
-There are two options to customize disks:
+#### nic
+default nic model to use. Likewise the disk implementation, it is recomended to pick default which ensure the maximum compatibility.
```
MarSik: Same as disk, the value is the name of nic that serves as a template.

workloads/templates/common-templates.md
```diff
@@ -134,43 +135,87 @@ in templates do not automatically affect VMs, or vice versa).
...
+kind: Template
+metadata:
+  name: Windows
+  annotations:
+    defaults.template.cnv.io/nic: e1000
+```
 
-#### Setting up name of PVC
-Each template has only one variable for disk customization - name of PersistentVolumeClaim(PVC).  Note: PVC should exist before you start your VM at first time. 
+#### volume
```
MarSik: Same, the value points to a template volume

workloads/templates/common-templates.md
```diff
@@ -134,43 +135,87 @@ in templates do not automatically affect VMs, or vice versa).
...
+kind: Template
+metadata:
+  name: Linux
+  annotations:
+    defaults.template.cnv.io/volume: custom-volume
+```
 
-### Editable fields related with disks
-Each template has ability to direct edit editable fields (see details and example above). By fact it provides same ability to choose name of PVC in comparison with changing variable
+#### network
```
MarSik: Same, points to a template network

fabiand:
So, I did a major change to the repo.
Once this PR is ready content wise, then I can do the update to the new layout.

If you want to do this feel free as well.

fromanirh:
@MarSik thanks for the comments, addressing them now
@fabiand if you can give me any pointer, I'll try the conversion myself

workloads/templates/common-templates.md
fromanirh: Interesting, I misunderstood https://github.com/kubevirt/common-templates/blob/master/templates/README.md, in particular """    # The goal of default disk is to define what kind of disk
    # is supported by the OS mainly in terms of bus (ide, scsi,
    # sata, virtio, ...)
    defaults.template.cnv.io/disk: default-disk"""

and, as you probably guessed, this error was replicated for all the other cases below. Solving them all.

fromanirh:
@MarSik @fabiand turns out the layout change *and* the transition to adoc are causing too many conflict, I can't make this PR mergeable again. So I sent https://github.com/kubevirt/user-guide/pull/177 starting from the last content here + comments from @MarSik addressed

fromanirh:
closing as per last comment

PR #167, opened 2018-12-06
Update sriov docs to not include explicit device requests

booxter:
With https://github.com/kubevirt/kubevirt/pull/1816 we no longer need to
fill in requests / limits for devices that implement sriov networks.
This patch updates the documentation to reflect that.


booxter:
ci test please

PR #171, opened 2018-12-07
Fix indocument links in disk-and-volumes doc

j-griffith:
Currently the links to sections of the doc are trying to use gh blob
references, but that doesn't actually work if you try and click on one
of the links.  The results is a 404.

Instead just use the GH MD syntax for internal doc links, that should
work and as as a bonus be portable for those that download the docs and are viewing
outside of GH or without net connectivity.


fabiand:
Thanks

PR #186, opened 2018-12-18
mention how to PXE-boot from an ovs-cni interface

dankenigsberg:
@fabiand are you looking for something else? I like the succinctness of this pr.

@SchSeba

### DIFF
--- a/creating-virtual-machines/interfaces-and-networks.adoc
+++ b/creating-virtual-machines/interfaces-and-networks.adoc
@@ -107,6 +107,7 @@ spec:
           bridge: {}
         - name: ovs-net
           bridge: {}
+          bootOrder: 1   # attempt to PXE-boot from this L2 networks
   networks:
   - name: default
     pod: {} # Stock pod network


dankenigsberg:
> Please also introduce a asection somewhere to describe the requirements on the network for PXE booting.
> Probbaly also worth noting that it does not work with the pod network.

I hoped that my "L2 network" was enough of a requirement. Even Pod network would work, if coincidentally it is backed by an L2 network that also has a dhcp server (not that I am aware of any such Pod network).

> Probably also worth noting that custom DHCP options can be passed to enable PXE booting for pod network - not?

I have no idea of such options, or even if dnsmasq can be a DHCP relay for the real PXE server. I find it very awkward to overload the default network with such a notion. @SchSeba do we have anything like this?

SchSeba:
Hi @dankenigsberg 
>I have no idea of such options, or even if dnsmasq can be a DHCP relay for the real PXE server. I find it very awkward to overload the default network with such a notion. @SchSeba do we have anything like this?

we can add this configuration to any interface

```golang
type DHCPOptions struct {
	// If specified will pass option 67 to interface's DHCP server
	// +optional
	BootFileName string `json:"bootFileName,omitempty"`
	// If specified will pass option 66 to interface's DHCP server
	// +optional
	TFTPServerName string `json:"tftpServerName,omitempty"`
}
```

So the virtual machine will get the ip address from the pod interface but also will be able to ask for a network installation.

fabiand:
Please add a section which uses words to decribe how PXE booting can be enabled.

Effectively just a paragraph describing the params which are added by the PR right now.

fabiand:
I'd still prefer to have a some text describing how pxe boot work,s but it's better than nothing ...

dankenigsberg:
> I'd still prefer to have a some text describing how pxe boot work,s but it's better than nothing ...

Sorry about not getting into it :-(

PR #191, opened 2019-01-14
Updated installation to use the operator

slintes:


fabiand:
Could we do a pre-release, say v0.14.0-alpha.0 to merge this patch?

PR #196, opened 2019-01-21
added tungstenfabric network setup

michaelhenkel:


dankenigsberg:
This contribution sounds most interesting. Since I am not familiar with TF, would you please elaborate here about how it integrates with Kubenetes and KubeVirt?

For example you have a CNI and a NetworkAttachmentDefinition, why cannot you use here the existing `multus` support? Can you have simple flannel handle the primary Pod network, but attach secondary network via Tungsten?

michaelhenkel:
@dankenigsberg When using TungstenFabric no other CNI can be used. Even the "Default Kubernetes Network" uses the TF CNI. i.e. When you do not specify a NIC and a network and the default POD network is used, TF CNI will be used.

dankenigsberg:
> When using TungstenFabric no other CNI can be used. 
Why is that? Assume that I have a cluster with TF as its default network. Why is it impossible to use Multus to pull in a secondary network from an unrelated CNI ?

> Even the "Default Kubernetes Network" uses the TF CNI. i.e. When you do not specify a NIC and a network and the default POD network is used, TF CNI will be used.

To pull the default/primary network into the VM, it should be enough to state `default`. Is there a problem why doing that if the default cni TF?

fabiand:
Just awaiting kubevirt/kubevirt#1956 to get merged

booxter:
@fabiand no it's not just a matter of waiting for that PR to merge. This documentation update is now completely invalid since 1) the other PR is no longer TF specific, 2) doesn't introduce a new network type. Instead, it introduces a new synonym for existing attribute.

fabiand:
Alright, we should still updat ethe guide to reflect this change.

dankenigsberg:
> Alright, we should still updat ethe guide to reflect this change.

not sure what change you are looking for. @booxter's https://github.com/kubevirt/kubevirt/pull/2107 makes `multus` support both Multus and TungstenFabric with zero change to users. @michaelhenkel  is no welcome to replace the `multus` string with something more neutral, and only then update this PR.

fabiand:
Oh okay, I misunderstood the scope then.
If there are no user facing changes - then there is obviously no change
required. Only if we change something wrt the defacto standard.

PR #203, opened 2019-02-15
Add multus as default pod network example.

greg-bock:
User guide update for https://github.com/kubevirt/kubevirt/pull/1924

### DIFF
--- a/creating-virtual-machines/interfaces-and-networks.adoc
+++ b/creating-virtual-machines/interfaces-and-networks.adoc
@@ -124,6 +124,15 @@ https://github.com/intel/multus-cni[Multus]. A version of multus after this
 https://github.com/intel/multus-cni/pull/174[Pull Request] is required
 (currently master).
 
+
+*Note the following:*
+
+* A multus default network and a pod network type are mutually exclusive.
+* The virt-launcher pod that starts the VMI will *not* have the pod network
+configured.
+* The multus delegate chosen as default *must* return at least one IP address.
+
+
 The following defines a VMI with an OVS network as the default.
 
 [source,yaml]
@@ -141,10 +150,6 @@ spec:
       default: true
       networkName: ovs-vlan-100
 ----
-
-*Note* that multus default network and a pod network type are mutually exclusive.
-The virt-launcher pod that starts the VMI will *not* have the pod network
-configured.
 
 genie
 ^^^^^
--- a/creating-virtual-machines/interfaces-and-networks.adoc
+++ b/creating-virtual-machines/interfaces-and-networks.adoc
@@ -140,21 +140,20 @@ Create a `NetworkAttachmentDefinition` with IPAM.
 apiVersion: "k8s.cni.cncf.io/v1"
 kind: NetworkAttachmentDefinition
 metadata:
-  name: ovs-vlan-200
+  name: macvlan-test
 spec:
   config: '{
-      "cniVersion": "0.3.1",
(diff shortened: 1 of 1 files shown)


fabiand:
@SchSeba can you please review?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -119,6 +119,26 @@ spec:
...
+spec:
+  domain:
+    devices:
+      interfaces:
+        - name: ovs-net
+          bridge: {}
+  networks:
+  - name: ovs-net
+    multus: # Multus network as default
+      default: true
```
SchSeba: Can you please add a comment to explain that with this configuration the virt-launcher pod will not get the pod network.

greg-bock: Done.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -119,6 +119,38 @@ spec:
...
+
+*Note the following:*
+
+* A multus default network and a pod network type are mutually exclusive.
+* The virt-launcher pod that starts the VMI will *not* have the pod network
+configured.
+* The multus delegate chosen as default *must* return at least one IP address.
+
+
+The following defines a VMI with an OVS network as the default.
```
phoracek: This example is little misleading since we did not implement IPAM in the OVS plugin.

You do not define any NetworkAttachmentDefinition in this example, could you please drop the OVS mention?

greg-bock: Sure

greg-bock: Added an OVS NAD with IPAM for the example.

phoracek: What I meant is, that kubevirt/ovs-cni plugin does not support IPAM, it is only L2. If you give it IPAM section, it either fails or ignores it.

Could you maybe use linux bridge instead https://github.com/containernetworking/plugins/tree/master/plugins/main/bridge ?

greg-bock: Sorry about that, I'll get right on it.

greg-bock: @phoracek how does it look now? (I've actually tested this one on a real setup to verify functionality).

phoracek:
LGTM! :)

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -119,6 +119,58 @@ spec:
...
+metadata:
+  name: macvlan-test
+spec:
+  config: '{
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "host-local",
+              "subnet": "10.250.250.0/24"
```
SchSeba: @greg-bock can you please take a look here https://github.com/containernetworking/plugins/tree/master/plugins/ipam/host-local

I think the `subnet` needs to by inside a `ranges` list

greg-bock: > Previous versions of the host-local allocator did not support the ranges property, and instead expected a single range on the top level. This is deprecated but still supported.

I wanted to keep the example as small as possible.

fabiand:
Yey.

PR #208, opened 2019-02-26
Added note about changed kubevirt account 

slintes:
Added note about changed kubevirt account which needs to be added to privileged SCC starting with Version 0.15.0.

I suggest another update once 0.15.0 is released, moving the new account into the normal documentation, and changing the note to mention the old account (or remove the note at all)

Signed-off-by: Marc Sluiter <msluiter@redhat.com>

### DIFF
--- a/administration/intro.adoc
+++ b/administration/intro.adoc
@@ -181,6 +181,9 @@ https://docs.openshift.com/container-platform/3.7/admin_guide/manage_scc.html[SC
 needs to be added prior KubeVirt deployment:
 
 - When you use the stable deployment flow:
+____________________________________
+Important: Starting with version v0.15.0 please replace `kubevirt-privileged` with `kubevirt-handler`
+____________________________________
 [source,bash]
 ----
 $ oc adm policy add-scc-to-user privileged -n kubevirt -z kubevirt-privileged


administration/intro.adoc
```diff
@@ -181,6 +181,9 @@ https://docs.openshift.com/container-platform/3.7/admin_guide/manage_scc.html[SC
 needs to be added prior KubeVirt deployment:
 
 - When you use the stable deployment flow:
+____________________________________
+Important: Starting with version v0.15.0 please replace `kubevirt-privileged` with `kubevirt-handler`
```
fabiand: What does replace mean in this case?

Put differently: Can you please update the complete document (or guide) to use this new service account?

slintes: replace that service account name in the next line
didn't want to update the original instruction yet, because 0.15.0 is only alpha released atm...?

fabiand: Ok, thanks for the clarification.

Now, what we'll do - I'll create a branch for each KubeVirt release to capture the documentation at that point in time.

Master will then always contain what will be in the next release.

The gap is that we need to publish the documentation for master including all other (or just the previous) release.

PR #220, opened 2019-03-21
Document how custom MAC address interacts with misc CNI plugins

booxter:
This update covers which plugins support `mac` attribute natively; which
work if combined with `tuning` plugin; and which are known to not work
at all. (The latter would require proper native support to be added in
the respective CNI code.)

### DIFF
--- a/creating-virtual-machines/interfaces-and-networks.adoc
+++ b/creating-virtual-machines/interfaces-and-networks.adoc
@@ -238,6 +238,54 @@ spec:
     pod: {}
 ----
 
+______________________________________________________________________________
+**Note:** If a specific MAC address is configured for a virtual machine
+interface, it's passed to the underlying CNI plugin that is expected to
+configure the backend to allow for this particular MAC address. Not every
+plugin has native support for custom MAC addresses.
+______________________________________________________________________________
+
+______________________________________________________________________________
+**Note:**  For some CNI plugins without native support for custom MAC
+addresses, there is a workaround, which is to use the `tuning` CNI plugin to
+adjust pod interface MAC address. This can be used as follows:
+
+[source,yaml]
+----
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: ptp-mac
+spec:
+  config: '{
+      "cniVersion": "0.3.1",
+      "name": "ptp-mac",
+      "plugins": [
+        {
+          "type": "ptp",
+          "ipam": {
+            "type": "host-local",
+            "subnet": "10.1.1.0/24"
+          }
+        },
+        {
+          "type": "tuning"
+        }
+      ]
+    }'
+----
+
+
+This approach may not work for all plugins. For example, OpenShift SDN is not
+compatible with `tuning` plugin.
+
+* Plugins that handle custom MAC addresses natively: `ovs`.
+* Plugins that are compatible with `tuning` plugin: `flannel`, `ptp`, `bridge`.
+* Plugins that don't need special MAC address treatment: `sriov` (in `vfio`
+  mode).
+______________________________________________________________________________
+
+
 Ports
 ^^^^^
 
--- a/creating-virtual-machines/interfaces-and-networks.adoc
+++ b/creating-virtual-machines/interfaces-and-networks.adoc
@@ -243,10 +243,10 @@ Ports
 
 Declare ports listen by the virtual machine
 
-__________________________________________________________________________________________________________
+______________________________________________________________________________
 *Note:* When using the slirp interface only the configured ports will be
 forwarded to the virtual machine.
-__________________________________________________________________________________________________________
+______________________________________________________________________________
 
 [cols=",,,",options="header",]
 |============================================
@@ -289,10 +289,11 @@ through a linux ``bridge''. The pod network IPv4 address is delegated to
 the virtual machine via DHCPv4. The virtual machine should be configured
 to use DHCP to acquire IPv4 addresses.
 
-________________________________________________________________________________________________________________________________________________________________________________________________________________________________
-**Note:** If a specific MAC address is not configured in the virtual machine interface spec
-the MAC address from the relevant pod interface is delegated to the virtual machine.
-________________________________________________________________________________________________________________________________________________________________________________________________________________________________
+______________________________________________________________________________
+**Note:** If a specific MAC address is not configured in the virtual machine
+interface spec the MAC address from the relevant pod interface is delegated to
+the virtual machine.
+______________________________________________________________________________
 
 
 [source,yaml]
@@ -312,12 +313,11 @@ spec:
 At this time, `bridge` mode doesn’t support additional configuration
 fields.
 
-________________________________________________________________________________________________________________________________________________________________________________________________________________________________
-*Note:* due to IPv4 address delagation, in `bridge` mode the pod doesn’t
-have an IP address configured, which may introduce issues with
-third-party solutions that may rely on it. For example, Istio may not
-work in this mode.
-________________________________________________________________________________________________________________________________________________________________________________________________________________________________
+______________________________________________________________________________
+*Note:* due to IPv4 address delagation, in `bridge` mode the pod doesn’t have
+an IP address configured, which may introduce issues with third-party solutions
+that may rely on it. For example, Istio may not work in this mode.
+______________________________________________________________________________
 
 slirp
 ^^^^^
@@ -343,10 +343,10 @@ spec:
 At this time, `slirp` mode doesn’t support additional configuration
 fields.
 
-_______________________________________________________________________________________________
-*Note:* in `slirp` mode, the only supported protocols are TCP and UDP.
-ICMP is _not_ supported.
-_______________________________________________________________________________________________
+______________________________________________________________________________
+*Note:* in `slirp` mode, the only supported protocols are TCP and UDP. ICMP is
+_not_ supported.
+______________________________________________________________________________
 
 More information about SLIRP mode can be found in
 https://wiki.qemu.org/Documentation/Networking#User_Networking_.28SLIRP.29[QEMU
@@ -361,7 +361,8 @@ IP addresses to virtual machines and hides them behind NAT.
 All the traffic exiting the virtual machine will be natted using the pod ip address.
 The virtual machine should be configured to use DHCP to acquire IPv4 addresses.
 
-To allow traffic gets into the virtual machine the port section need to be configured in the interface spec.
+To allow traffic gets into the virtual machine the port section need to be
+configured in the interface spec.
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


booxter:
@phoracek @myakove could you please review this PR makes sense?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -238,15 +238,68 @@ spec:
...
+spec:
+  config: '{
+      "cniVersion": "0.3.1",
+      "name": "ptp-mac",
+      "plugins": [
+        {
+          "type": "ptp",
+          "ipam": {
+            "type": "host-local",
+            "subnet": "10.1.1.0/24"
```
myakove: Since it's all about MAC maybe add mac address to the example yaml?

booxter: @myakove which example should I add it to? From kubevirt perspective, I would need to add a separate example with macAddress attribute. The MAC address is not part of network CRD. Do you want me to update the VM spec defined in line 223+?

myakove:
If the MAC is not part of the crd then you can ignore my comment. I thought that we configure the Mac in the NetworkAttacmetDefinition.

booxter:
@myakove one *can* configure it in the CRD, in which case same MAC address is used for all ports (not very useful); but in our setup MAC address is generated by macpool / configured in macAddress in VMI spec and then passed to CNI plugin (multus) that then proxies it to each of sub-plugins (e.g. bridge and tuning).

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -238,15 +238,68 @@ spec:
     pod: {}
 ----
 
+______________________________________________________________________________
+**Note:** If a specific MAC address is configured in the virtual machine
+interface spec, it's passed to underlying CNI plugin that should configure
+backend to allow for this particular address. While some plugins are known to
+work without special support for custom MAC addresses, others do (for example,
```
phoracek: "others do**n't**"?

The OpenShiftSDN example is maybe too specific. Do you think it would be worth saying that primary network plugins (calico, flannel, openshiftSDN) does not support it? However, then we should double check that they indeed don't.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -238,15 +238,68 @@ spec:
...
+OpenShift SDN needs to configure OVS flows to allow traffic meant for the
+chosen MAC address).
+______________________________________________________________________________
+
+______________________________________________________________________________
+**Note:** If a plugin doesn't support `mac` attribute defined by Kubernetes
```
phoracek: This is confusing. In the previous section you say some are know to work without special effort. @SchSeba could correct me here, but I think that most plugins are OK with custom address (unlike OpenShiftSDN), but the address must be set using the tuning CNI.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -238,15 +238,68 @@ spec:
...
+      ]
+    }'
+----
+
+
+This approach won't work for other plugins that needs special treatment for
+custom MAC addresses. For example, OpenShift SDN is not compatible with
+`tuning` plugin.
+
+Plugins that handle custom MAC addresses natively: `ovs`.
```
phoracek: We do? I think you need to use tuning with OVS CNI.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -238,15 +238,68 @@ spec:
...
+
+Plugins that don't need special MAC address treatment: `sriov`.
```
phoracek: That means sriov can read "mac" parameter on its own or do we need to use tuning?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -343,10 +396,10 @@ spec:
 At this time, `slirp` mode doesn’t support additional configuration
 fields.
 
-_______________________________________________________________________________________________
-*Note:* in `slirp` mode, the only supported protocols are TCP and UDP.
-ICMP is _not_ supported.
-_______________________________________________________________________________________________
+______________________________________________________________________________
+*Note:* in `slirp` mode, the only supported protocols are TCP and UDP.  ICMP is
```
phoracek: Noooo, two spaces before ICMP

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -379,10 +433,11 @@ spec:
     pod: {}
 ----
 
-_______________________________________________________________________________________________
+______________________________________________________________________________
 *Note:* Masquerade is only allow to be connected to a pod network backend.
-*Note:* The network CIDR for can be configured in the pod network section using `vmNetworkCIDR`.
-_______________________________________________________________________________________________
+*Note:* The network CIDR for can be configured in the pod network section using
```
phoracek: CIDR for what?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -474,10 +528,10 @@ spec:
       networkName: sriov-net-crd
 ----
 
-__________________________________________________________________________________
-*Note:* you need to enable the SRIOV feature gate to use the feature.
-For example:
-__________________________________________________________________________________
+______________________________________________________________________________
+*Note:* you need to enable the SRIOV feature gate to use the feature.  For
```
phoracek: Two spaces :(

creating-virtual-machines/interfaces-and-networks.adoc
booxter: Don't make me fix all the typos and errors in the document pretty please. :)

creating-virtual-machines/interfaces-and-networks.adoc
booxter: ...which is bad because?..

creating-virtual-machines/interfaces-and-networks.adoc
booxter: Explain why it's bad.

creating-virtual-machines/interfaces-and-networks.adoc
booxter: For SR-IOV devices, since we use VFIO, we don't care b/c there is no netlink device once we pass it to guest.

creating-virtual-machines/interfaces-and-networks.adoc
booxter: @phoracek what is this code for then? https://github.com/kubevirt/ovs-cni/blob/50c2abbc4209ef07fcce8fb41c044c24df571690/pkg/plugin/plugin.go#L135-L154

creating-virtual-machines/interfaces-and-networks.adoc
booxter: Plugin not handling `mac` attribute natively == plugin doesn't support it. That you can then work it around with `tuning` plugin is just a function of the original plugin relying on netlink devices only, or not. I will think how to reword it.

creating-virtual-machines/interfaces-and-networks.adoc
booxter: I will fix the typo, thanks.

I haven't checked whether `calico` or `flannel` support `mac` attribute. I suspect they don't do it natively but since they are netlink based they should support it through `tuning` plugin.

phoracek: The problem would not be that `tuning` does not change the MAC, but that it would break the connection (similar to OpenShiftSDN). I think it would be valuable for users to get simple **supported/not supported** list. If you think it is not a lot of work.

creating-virtual-machines/interfaces-and-networks.adoc
phoracek: It's not consistent across the whole document. If you explain to me a double-space rule that would apply to the whole document, I'm willing to stop nit-picking.

creating-virtual-machines/interfaces-and-networks.adoc
phoracek: :man_facepalming: forgot about this one

booxter: it's ok, it was 3 months ago :)

creating-virtual-machines/interfaces-and-networks.adoc
booxter: It's just what vim's re-wrapping to X chars does this by default. Believe me I haven't put these spaces in manually.

I will remove them but I find it's weird that they get nit-picked while we tend to ignore commit messages, comments, wrapping, copy paste in other places. ;)

phoracek: Never surrender! It's simple, it is easy to point out and suggest solution for double space, not so for a commit message (hidden by evil GitHub). But you are right, we should step up the nit-picking game.

creating-virtual-machines/interfaces-and-networks.adoc
booxter: @phoracek do we even test with `calico`? How do we define the list of plugins we would like to list here? There are tens of different plugins. Should we document each of them?

For `flannel`, I guess it makes sense to have it documented because it's quite popular.

phoracek: You are right. Let's do it just for Flannel which is tested under kubevirt.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -238,15 +238,68 @@ spec:
...
+      "plugins": [
+        {
+          "type": "ptp",
+          "ipam": {
+            "type": "host-local",
+            "subnet": "10.1.1.0/24"
+          },
+        },
+        {
+          "type": "tuning",
```
SchSeba: please remove the `,` this is invalid in json.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -238,15 +238,68 @@ spec:
...
+  config: '{
+      "cniVersion": "0.3.1",
+      "name": "ptp-mac",
+      "plugins": [
+        {
+          "type": "ptp",
+          "ipam": {
+            "type": "host-local",
+            "subnet": "10.1.1.0/24"
+          },
```
SchSeba: please remove the `,` this is invalid in json.

creating-virtual-machines/interfaces-and-networks.adoc
booxter: I've checked that `flannel` works with `tuning` plugin. I've changed wording here somewhat; there is an explicit list of plugins, classified by their MAC support status, below this section. I hope it's better now.

creating-virtual-machines/interfaces-and-networks.adoc
booxter: Mentioned I assume VFIO mode here (which is the mode we should use for kubevirt).

creating-virtual-machines/interfaces-and-networks.adoc
booxter: Removed two spaces. The Earth is saved!

creating-virtual-machines/interfaces-and-networks.adoc
booxter: Removed two spaces. Batman can rest now!

creating-virtual-machines/interfaces-and-networks.adoc
booxter: It was not malicious but I missed this fix from the upload. But now I am lazy enough to not bother and play the 'not my patch' card.

booxter:
@SchSeba @phoracek this update should do. Hopefully I addressed every comment, even if not satisfied.

PR #227, opened 2019-04-02
Document live migration on node-drain

rmohr:
* Document cluster-wide migration limits and how to change them
* Document `evictionPolicy` for live-migrate on drains

### DIFF
--- a/administration/node-eviction.adoc
+++ b/administration/node-eviction.adoc
@@ -71,11 +71,7 @@ if the `kubevirt.io/drain:NoSchedule` taint is added to a nodes:
 [source,yaml]
 ----
 spec:
-  evictionPolicy:
-    taints:
-      - effect: NoSchedule
-        key: kubevirt.io/drain
-        strategy: LiveMigrate
+  evictionStrategy: LiveMigrate
 ----
 
 Here a full VMI:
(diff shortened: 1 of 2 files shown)


administration/node-eviction.adoc
```diff
@@ -60,6 +60,80 @@ target node.
...
 
+How to evacuate VMIs via Live Migration from a Node
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If the *LiveMigration* feature gate is enabled, it is possible to specify an
+`evictionStrategy` on VMIs which will react with live-migrations on specific
+taints on nodes. The following snipped on a VMI ensures that the VMI is migrated
+if the `kubevirt.io/drain:NoSchedule` taint is added to a nodes:
+
+[source,yaml]
```
fabiand: Please add an example of how a evictionPolicy needs to look in order to tie in with the regular `kubectl drain`

rmohr: For completeness that would be setting the config to

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubevirt-config
  namespace: kubevirt
  labels:
    kubevirt.io: ""
data:
  feature-gates: "LiveMigration"
  migrations: |-
    nodeDrainTaintKey: nodes.kubernetes.io/noSchedule
```

However, we decided to not document that, because the implicit `kubectl cordon` is the reason why this taint gets added and you normally don't want migrations if a node gets marked with `NoSchedule` via `kubectl cordon`.

rmohr: had a copy-pase error. It is `node.kubernetes.io/unschedulable`

PR #237, opened 2019-04-26
Document migration and cancel migration in the UI

jelkosz:


jelkosz:
@pcbailey please review

web-interface/controlling-vms-modules/migrate-vm-web.adoc
```diff
@@ -0,0 +1,12 @@
+[[migrate-vm-web]]
+== Live Migrating a Virtual Machine
```
pcbailey: "Live Migrating" sounds odd to me, but if that's the phrase being used, I guess we'll go with it; otherwise, I'd suggest "Migrating a Running Virtual Machine". Similar changes would be necessary in the following section if that change is made, though.

web-interface/resource-modules/vm-actions-web.adoc
```diff
@@ -19,5 +19,13 @@
...
 |_All_
 |Permanently delete the virtual machine from the cluster.
+
+|*Migrate Virtual Machine*
+|_Running_
+|Live migrates the virtual machine to an another node.
+
+|*Cancel Virtual Machine Migration*
+|_Migrating_
+|Cancells the migration of the virtual machine.
```
pcbailey: typo: "Cancells" should be "Cancels"

web-interface/controlling-vms-modules/migrate-vm-web.adoc
jelkosz: It is called the "live migration" so anyone looking for live migration would look for something like this, so I would stay with it.

web-interface/resource-modules/vm-actions-web.adoc
jelkosz: done

jelkosz:
@fabiand can you please merge?

PR #238, opened 2019-04-29
Update Containerized Data Importer version to 1.8

olberger:
Reusing instructions from Containerized Data Importer's README


fabiand:
Thanks!

PR #260, opened 2019-06-19
Document cluster memory overcommit

ahadas:
Document [#2304](https://github.com/kubevirt/kubevirt/pull/2304)

### DIFF
--- a/using-virtual-machines/overcommit.adoc
+++ b/using-virtual-machines/overcommit.adoc
@@ -47,7 +47,7 @@ spec:
   terminationGracePeriodSeconds: 30
   domain:
     resources:
-    overcommitGuestOverhead: true
+      overcommitGuestOverhead: true
       requests:
         memory: 1024M
 [...]
@@ -81,7 +81,7 @@ spec:
   terminationGracePeriodSeconds: 30
   domain:
     resources:
-    overcommitGuestOverhead: true
+      overcommitGuestOverhead: true
       requests:
         memory: 1024M
     memory:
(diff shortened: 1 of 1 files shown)


fabiand:
@MarSik thoughts?

MarSik:
The docs look good to me. The config option is just a simplification for users, right? Because you always could override the value in spec.resources. Also I wonder whether we are accounting for the qemu and guest overhead here. Because for normal cases (100%) the resources should not match the VM, but it should be VM + expected overhead (vgpu buffers, memory tables, ..).

ahadas:
> The docs look good to me. The config option is just a simplification for users, right? Because you always could override the value in spec.resources. Also I wonder whether we are accounting for the qemu and guest overhead here. Because for normal cases (100%) the resources should not match the VM, but it should be VM + expected overhead (vgpu buffers, memory tables, ..).

Well, it simplifies things but the main benefit is that it replaces the ridiculously low default memory request of 8M with something that makes more sense. Users can still exlicitely specify everything if they wish, yes. And regarding the overhead, its overcommit is specified diffetently so no, it's not affected by this

MarSik:
Then I would actually put itno the docs an explanation how this affects the actual memory request that will be used for the pod. It should be VM memory / overcommit + overhead, right?

ahadas:
> Then I would actually put itno the docs an explanation how this affects the actual memory request that will be used for the pod. It should be VM memory / overcommit + overhead, right?

Right, unless [OvercommitGuestOverhead](https://github.com/kubevirt/kubevirt/pull/2416) is set.
I would rather prefer not documenting the existing configuration of the overhead as part of this PR, although it is indeed related, since this PR doesn't change it and in light of the [work-in-progress in that area](https://github.com/kubevirt/kubevirt/pull/2406).

MarSik:
@ahadas Then just maybe add a line saying something like: The actual mempory request depends on additional configuration options like OvercommitGuestOverhead.

ahadas:
> @ahadas Then just maybe add a line saying something like: The actual mempory request depends on additional configuration options like OvercommitGuestOverhead.

Done

PR #266, opened 2019-07-29
Change examples with bridge to masquerade

alonSadan:
in this PR we changed the example vm configuration
which had bridge as the default
binding method to masqurade.
That's regarding the issues discussed in detail here:
https://github.com/kubevirt/kubevirt/pull/2537
and here: https://github.com/kubevirt/kubevirt/pull/2493
We would like the examples to be consistent with
the changes suggested in those two PRs, and prevent from
a random user to copy-paste the examples and
receive an unexpected behavior.

### DIFF
--- a/creating-virtual-machines/interfaces-and-networks.adoc
+++ b/creating-virtual-machines/interfaces-and-networks.adoc
@@ -54,7 +54,7 @@ spec:
     devices:
       interfaces:
         - name: default
-          bridge: {}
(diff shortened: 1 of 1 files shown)


alonSadan:
@SchSeba can you please review ?

SchSeba:
/cc @rmohr 

can you please take a look on this PR ? :)

PR #267, opened 2019-07-29
Note and describe the option to disable bridge

alonSadan:
- A note was added, with example, on the flag that
allow us to disable and enable bridge interface
on pod network.

- A note was added suggesting the user to configure
masquerade as a default interface.

regards changes in this PR: 
https://github.com/kubevirt/kubevirt/pull/2493

### DIFF
--- a/creating-virtual-machines/interfaces-and-networks.adoc
+++ b/creating-virtual-machines/interfaces-and-networks.adoc
@@ -409,15 +409,45 @@ spec:
   - name: red
     multus:
       networkName: red
------
+----
 
 At this time, `bridge` mode doesn’t support additional configuration
 fields.
 
 ______________________________________________________________________________
-*Note:* due to IPv4 address delagation, in `bridge` mode the pod doesn’t have
+*Note:* due to IPv4 address delegation, in `bridge` mode the pod doesn’t have
 an IP address configured, which may introduce issues with third-party solutions
 that may rely on it. For example, Istio may not work in this mode.
+______________________________________________________________________________
+
+______________________________________________________________________________
+*Note:* admin can forbid using `bridge` interface type for pod networks via a 
+designated configuration flag. 
+To achieve it, the admin should set the following option to `false`:
+______________________________________________________________________________
+
+
+[source,yaml]
+-----
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubevirt-config
+  namespace: kubevirt
+  labels:
+    kubevirt.io: ""
+data:
+  permitBridgeInterfaceOnPodNetwork: "false"
+-----  
+
+______________________________________________________________________________
+*Note:* binding the pod network using `bridge` interface type may cause issues.
+Other than the third-party issue mentioned in the above note, 
+live migration is not allowed with a pod network binding of `bridge` interface type,
+and also some CNI plugins might not allow to use a custom MAC address for your VM instances.
+If you think you may be affected by any of issues mentioned above, 
+consider changing the default interface type to `masquerade`, 
+and disabling the `bridge` type for pod network, as shown in the example above.
 ______________________________________________________________________________
 
 slirp
@@ -482,6 +512,9 @@ spec:
 
 ______________________________________________________________________________
 *Note:* Masquerade is only allowed to connect to the pod network.
+______________________________________________________________________________
+
+______________________________________________________________________________
 *Note:* The network CIDR can be configured in the pod network section using the
 `vmNetworkCIDR` attribute.
 ______________________________________________________________________________


alonSadan:
@SchSeba  and here to... Thanks:)

alonSadan:
@slintes what do you think?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -417,6 +417,22 @@ ______________________________________________________________________________
 *Note:* due to IPv4 address delagation, in `bridge` mode the pod doesn’t have
 an IP address configured, which may introduce issues with third-party solutions
 that may rely on it. For example, Istio may not work in this mode.
+*Note:* there is an option to dissable bridge interface with pod network.
+This is done with a deignated flag.
```
slintes: typo: designated

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -417,6 +417,22 @@ ______________________________________________________________________________
...
+The default value of the flag is 'true', meaning bridge interface is enabled.
+to disable it cahnge flag value to false. For example:
```
slintes: typo: change

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -417,6 +417,22 @@ ______________________________________________________________________________
...
+______________________________________________________________________________
+apiVersion: v1
```
slintes: formatting is broken, you can check it by clicking "view file" in the menu on the top right.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -419,6 +419,25 @@ an IP address configured, which may introduce issues with third-party solutions
 that may rely on it. For example, Istio may not work in this mode.
 ______________________________________________________________________________
 
+______________________________________________________________________________
+*Note:* there is an option to dissable bridge interface with pod network.
```
booxter: disable (one s)

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -419,6 +419,25 @@ an IP address configured, which may introduce issues with third-party solutions
...
+This is done with a designated flag.
+The default value of the flag is 'true', meaning bridge interface is enabled.
+to disable it change flag value to false. For example:
```
booxter: this line should start with upper case

booxter: you single-quote 'true' above but not false here. Please make the text consistent.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -419,6 +419,25 @@ an IP address configured, which may introduce issues with third-party solutions
 that may rely on it. For example, Istio may not work in this mode.
 ______________________________________________________________________________
 
+______________________________________________________________________________
```
booxter: as an uninitiated reader, the very first question I'll have is: why would I want to avoid bridge mode? Is it because of the other note above about third-party solutions? No other reasons? (There are other reasons though, specifically, migration issues as well as issues with OpenShift that I, as a reader, would probably appreciate to know about to make the right decision.) So should we document these other reasons to make the new text more helpful?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -419,6 +419,25 @@ an IP address configured, which may introduce issues with third-party solutions
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubevirt-config
+  namespace: kube-system
+  labels:
+    kubevirt.io: ""
+data:
+ permitBridgeInterfaceOnPodNetwork: "false"
```
booxter: looks like there is a single indentation space in this line but other lines use two spaces

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -419,6 +419,25 @@ an IP address configured, which may introduce issues with third-party solutions
...
+______________________________________________________________________________
+*Note:* consider changing the default interface in the config map to masquerade.
```
booxter: why should I though? I'm still not sure if I'm going to be affected by any issues if I don't use third party solutions like Istio.

alonSadan: Thanks @booxter . you are right, that is a bit misleading.
I added some example reasons why one would prefer to not use bridge, and a general concern about to bridge binding to reason the recommendation to use masquerade. 
+  making the explanation about the third party issue and the recommendation to use masquerade in one note, or keep as it is - separate?   
What do you think?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -414,11 +414,36 @@ At this time, `bridge` mode doesn’t support additional configuration
...
 that may rely on it. For example, Istio may not work in this mode.
 ______________________________________________________________________________
 
+______________________________________________________________________________
+*Note:* there is an option to disable bridge interface with pod network.
+This is done with a designated flag.
+The default value of the flag is 'true', meaning bridge interface is enabled.
+To disable it change flag value to 'false'. For example:
+______________________________________________________________________________
+apiVersion: v1
```
SchSeba: you can do something like this https://github.com/kubevirt/user-guide/pull/267/files#diff-5ac883b57d09207b7ff7a1dae3413f18L399 so the code will look better

creating-virtual-machines/interfaces-and-networks.adoc
booxter: I don't really care, both are ok in my view.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -415,11 +415,40 @@ At this time, `bridge` mode doesn’t support additional configuration
...
+  namespace: kube-system
+  labels:
+    kubevirt.io: ""
+data:
+  permitBridgeInterfaceOnPodNetwork: "false"
+-----  
+______________________________________________________________________________
+*Note:* other than the third-party issue mentioned in the above note, binding 
+with bridge may cause other unexpected behaviors. For example, 
+live - migration is not allowed with bridge binding and also some
```
booxter: dash between `live` and `migration` should be removed; also, add a comma before `and`

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -415,11 +415,40 @@ At this time, `bridge` mode doesn’t support additional configuration
...
+CNI plugins might not allow to use a costume mac for your VM instances.
```
booxter: s/costume/custom/
s/mac/MAC address/

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -415,11 +415,40 @@ At this time, `bridge` mode doesn’t support additional configuration
...
+For those example issues and the fact that more bugs caused by the IP
```
booxter: This sentence would appreciate some love, for example: "If you think you may be affected by any of issues mentioned above, consider changing the default interface type to `masquerade` and disabling the `bridge` type for pod interfaces, as shown in the example above."

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -575,6 +607,23 @@ spec:
       networkName: sriov-net-crd
 ----
 
+______________________________________________________________________________
+*Note:* you need to enable the SRIOV feature gate to use the feature. For
```
booxter: This note is probably a spurious remnant of a rebase. I don't think we document the feature gate anymore considering that it's enabled by default. Please remove the note.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,48 @@ spec:
...
 
 ______________________________________________________________________________
-*Note:* due to IPv4 address delagation, in `bridge` mode the pod doesn’t have
+*Note:* due to IPv4 address delegation, in `bridge` mode the pod doesn’t have
 an IP address configured, which may introduce issues with third-party solutions
 that may rely on it. For example, Istio may not work in this mode.
 ______________________________________________________________________________
 
+______________________________________________________________________________
+*Note:* there is an option to disable bridge interface with pod network.
```
booxter: I am sorry for not catching it yesterday (in my defense, it was too late and I was sleepy already) but: this first sentence is not very clear. Specifically, one could read it as: "There is an option that allows a user/admin to disable adding a pod interface to a VMI (that is usually/by default of bridge type)." I think it would be great if we can reword it to remove this ambiguity. For example,

```
Note: Admin can forbid using `bridge` interface type for pod networks via a designated configuration flag. To achieve it, the admin should set the following option to `true`:
...
```

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,48 @@ spec:
...
+  name: kubevirt-config
+  namespace: kube-system
+  labels:
+    kubevirt.io: ""
+data:
+  permitBridgeInterfaceOnPodNetwork: "false"
+-----  
+
+______________________________________________________________________________
+*Note:* other than the third-party issue mentioned in the above note, binding
```
booxter: let's be absolutely clear here too, by saying that binding *the pod network* using `bridge` interface type may cause issues.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,48 @@ spec:
...
+  labels:
+    kubevirt.io: ""
+data:
+  permitBridgeInterfaceOnPodNetwork: "false"
+-----  
+
+______________________________________________________________________________
+*Note:* other than the third-party issue mentioned in the above note, binding 
+with bridge may cause other unexpected behaviors. For example, 
+live-migration is not allowed with bridge binding, and also some
```
booxter: I don't think it's a single word "live-migration" but that it's two words "live migration"

booxter: again, let's stress we are talking specifically about pod network here: "For example live migration is not allowed with a pod network binding of `bridge` interface type..."

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,48 @@ spec:
...
+CNI plugins might not allow to use a custom MAC for your VM instances.
```
booxter: MAC -> MAC *address*

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,48 @@ spec:
...
+If you think you may be affected by any of issues mentioned above, 
+consider changing the default interface type to masquerade,
```
booxter: Consider quoting keywords like `masquerade` and `bridge` with back-quotes so that they are marked in the documentation as in this comment.

Gemfile
```diff
@@ -0,0 +1,6 @@
+Encoding.default_external = Encoding::UTF_8
```
booxter: hm why this file added?

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,47 @@ spec:
...
 
 ______________________________________________________________________________
-*Note:* due to IPv4 address delagation, in `bridge` mode the pod doesn’t have
+*Note:* due to IPv4 address delegation, in `bridge` mode the pod doesn’t have
 an IP address configured, which may introduce issues with third-party solutions
 that may rely on it. For example, Istio may not work in this mode.
 ______________________________________________________________________________
 
+______________________________________________________________________________
+*Note:* admin can forbid using `bridge` interface type for `pod networks` via a
```
booxter: it's probably not a good idea to "quote" pod networks since it's not a literal from any spec or API but just an entity.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,47 @@ spec:
...
+______________________________________________________________________________
+
+
+[source,yaml]
+-----
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubevirt-config
+  namespace: kube-system
```
booxter: (sorry) do we still use kube-system? I thought we switched to a different one. I suggest you trim this completely from the example like we do for other examples.

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,47 @@ spec:
...
+  labels:
+    kubevirt.io: ""
+data:
+  permitBridgeInterfaceOnPodNetwork: "false"
+-----  
+
+______________________________________________________________________________
+*Note:* binding the `pod network` using `bridge` interface type may cause issues.
```
booxter: same about not quoting `pod network`

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,47 @@ spec:
...
+Other than the third-party issue mentioned in the above note, 
+live migration is not allowed with a `pod network` binding of `bridge` interface type,
```
booxter: ditto

creating-virtual-machines/interfaces-and-networks.adoc
```diff
@@ -409,17 +409,47 @@ spec:
...
+and also some CNI plugins might not allow to use a custom MAC address for your VM instances.
+If you think you may be affected by any of issues mentioned above, 
+consider changing the default interface type to `masquerade`, 
+and disabling the `bridge` type for `pod network`, as shown in the example above.
```
booxter: ditto

Gemfile
alonSadan: please see : 
https://github.com/kubevirt/user-guide/pull/285
It did not passed CI without it...

iranzo: @alonSadan the #285 has been merged, can you rebase on top of current master?

 Thanks!

iranzo:
@alonSadan  Commits must be signed (git commit -s)

To update this one, execute git commit --amend -s to update the commit message and then force push again

Thanks!

creating-virtual-machines/interfaces-and-networks.adoc
alonSadan: Thanks @booxter .You are right, the namespace is "kubevirt".
I saw that most of the configMap examples spread across the user-guide do include namespace, even though in this file most of the examples (not configMap examples) don't.
would you prefer to trim namespace or update it to "kubevirt"?

booxter: Either way, as long as you don't put the obsolete name into examples. Since you may want to clean up other examples by e.g. adding the right namespace name to them, if I were you, I would leave the namespace out of new examples in this PR and then send a follow-up PR that adds correct namespaces to all examples in this file. But that's after this PR lands.

alonSadan: The thing is that config-map has to be namespaced under kubevirt namespace so the configMap example should include kubevirt-namespace ,while VMs and NADs can also use the default namespce so the examples can also not include namespace.
what do you think @booxter ?

booxter:
I believe Marc's requests were handled already.

PR #270, opened 2019-08-01
Remove info about sriov feature gate

fedepaol:
With https://github.com/kubevirt/kubevirt/pull/2546 sriov is enabled by default and there is no need of setting the feature gate. This PR removes the instructions about setting the gate.


slintes:
Thanks. Let's wait with merging until v0.20.0 is released, in order to avoid confusion.

/hold

PR #273, opened 2019-08-02
containerDisks on scratch

rmohr:
Document that containerdisks can be based on `scratch`, starting with kubevirt 0.20.

### DIFF
--- a/creating-virtual-machines/disks-and-volumes.adoc
+++ b/creating-virtual-machines/disks-and-volumes.adoc
@@ -415,22 +415,23 @@ incompatible with Registry Disks.
 containerDisk Workflow Example
 ++++++++++++++++++++++++++++++
 
-Users push VM disks into the container registry using a KubeVirt base
-image designed to work with the Registry Disk feature. The latest base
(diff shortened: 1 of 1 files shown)


creating-virtual-machines/disks-and-volumes.adoc
```diff
@@ -415,22 +415,23 @@ incompatible with Registry Disks.
...
-Users push VM disks into the container registry using a KubeVirt base
-image designed to work with the Registry Disk feature. The latest base
-container image is *kubevirt/container-disk-v1alpha*.
-
-Using this base image, users can inject a VirtualMachineInstance disk
-into a container image in a way that is consumable by the KubeVirt
-runtime. Disks placed into the base container must be placed into the
-`/disk` directory. Raw and qcow2 formats are supported. Qcow2 is
-recommended in order to reduce the container image’s size.
+Using this base image, users can inject a VirtualMachineInstance disk into a
```
slintes: "this base image" does not make much sense anymore after removing the prior paragraph about the old container-disk image?

rmohr: true

PR #277, opened 2019-08-14
State that only N-1 to N is supported

### LINKED ISSUE #276: Upgrade docs state N-1 to N
We should *STATE* that only N-1 to N is supported in big red bold text :)

https://kubevirt.io/user-guide/docs/latest/administration/intro.html#update

iranzo:
Closes  #276

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,3 +20,20 @@ spellcheck:
 
 test:
 	test "$$(make -s find-unref) -eq 0"
+
+netlify: clean
+	git checkout -b current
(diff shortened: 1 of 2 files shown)


iranzo:
image

https://deploy-preview-277--kubevirt-user-guide.netlify.com/latest/administration/intro.html

iranzo:
It's based on #290 code so that it renders properly like above URL

iranzo:
No, as it was based on the latest, it's just adding a commit on top... :-)

PR #281, opened 2019-08-21
Recommend OpenShift SR-IOV operator for deployment

booxter:


booxter:
@fedepaol feel free to expand, though yes, my intention is to avoid details because things may change in the future in their repo and we probably don't want to update our docs every time it happens.

fedepaol:
Your approach definitely sounds better, since it's immune to changes in the operator behaviour and I added the upstream specific bits in the operator's quickstart guide.

Let's keep it this way.

booxter:
@slintes perhaps you could nudge it in merge queue. Thanks!

booxter:
I'll take a risk and merge this myself. It has two approvals from networking team members, and it doesn't seem controversial, just lacks attention of usual suspects to merge patches. If it's not ok, let me know and I will refrain from doing it in the future.

slintes:
> just lacks attention of usual suspects to merge patches

since you pinged me: I was on PTO

PR #282, opened 2019-08-22
Fix add path for container disk

petrkotas:
The Add have to end with / to copy into
folder, otherwise it will create file.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/user-guide/282)


slintes:
good catch!
that might even explain why I had problems with my own containerdisk on my test env, while fixing virt-handler. Didn't take the time to investigate it, because published containerdisks worked.

/lgtm
/approve

PR #287, opened 2019-09-14
Added script to delete all objects within a ns

mjschmidt:
Added script to delete all objects within a namespace to kubevirt intro documentation guide.

### DIFF
--- a/administration/intro.adoc
+++ b/administration/intro.adoc
@@ -332,4 +332,55 @@ to fix it, delete manually finalizer from the resource.
 ----
 $ kubectl -n kubevirt patch kv kubevirt --type=json -p '[{ "op": "remove", "path": "/metadata/finalizers" }]'
 ----
+
+Note: If you find your namespace is still hanging in terminationstate this script will delete all objects inside of a namespace.
+[source,bash]
+```
+# the namespace of kubevirt installation
+export namespace=kubevirt
+export labels=("operator.kubevirt.io" "operator.cdi.kubevirt.io" "kubevirt.io" "cdi.kubevirt.io")
+export namespaces=(default ${namespace} "<other-namespaces that has kubevirst resources>")
+
+kubectl -n ${namespace} delete kv kubevirt
+kubectl -n ${namespace} patch kv kubevirt --type=json -p '[{ "op": "remove", "path": "/metadata/finalizers" }]'
+kubectl get vmis --all-namespaces -o=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace,FINALIZERS:.metadata.finalizers --no-headers | grep foregroundDeleteVirtualMachine | while read p; do
+    arr=($p)
+    name="${arr[0]}"
+    namespace="${arr[1]}"
+    kubectl patch vmi $name -n $namespace --type=json -p '[{ "op": "remove", "path": "/metadata/finalizers" }]'
+done
+
+for i in ${namespaces[@]}; do
+    for label in ${labels[@]}; do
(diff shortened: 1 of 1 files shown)


iranzo:
@mjschmidt Commits must be signed (git commit -s)

To update this one, execute git commit --amend -s to update the commit message and then force push again 

Thanks!

mjschmidt:
okay probably get to this friday

administration/intro.adoc
```diff
@@ -332,4 +332,55 @@ to fix it, delete manually finalizer from the resource.
 ----
 $ kubectl -n kubevirt patch kv kubevirt --type=json -p '[{ "op": "remove", "path": "/metadata/finalizers" }]'
 ----
+
+Note: If you find your namespace is still hanging in terminationstate this script will delete all objects inside of a namespace.
```
dankenigsberg: this does not sound very safe. can you add a note why this is needed? how long should we typically wait for the normal cleanup to finish?

administration/intro.adoc
```diff
@@ -332,4 +332,55 @@ to fix it, delete manually finalizer from the resource.
...
+kubectl get vmis --all-namespaces -o=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace,FINALIZERS:.metadata.finalizers --no-headers | grep foregroundDeleteVirtualMachine | while read p; do
+    arr=($p)
+    name="${arr[0]}"
+    namespace="${arr[1]}"
+    kubectl patch vmi $name -n $namespace --type=json -p '[{ "op": "remove", "path": "/metadata/finalizers" }]'
+done
+
+for i in ${namespaces[@]}; do
+    for label in ${labels[@]}; do
+        kubectl -n ${i} delete deployment -l ${label}
```
dankenigsberg: this is quite repetitive. I think that
```
for kind in deployment ds rs ....; do
```
would improve this. However, this code is tightly related to what hco does; I would suspect this code should sit in hco and not here.

mjschmidt: It is repetitive. The point is to get every single possible kubernetes object

administration/intro.adoc
mjschmidt: This is needed to clean up all kubernetes objects when uninstall does not go as expected. This usually happens when things are not cleaned up in the correct order, or race conditions happen.

mjschmidt:
I forgot about this. Will do it later today

mjschmidt:
Hi Kubevirt folks, I cannot sign this because I deleted the repo from my personal github a few weeks ago not realizing this was not pulled.

I can create a new pull request if desired and make the changes again on a new branch.

dankenigsberg:
Well, I personally do not understand the problem you try to solve by doc. I wish you tell what typically goes wrong, and if like to see if we can fix it programmatically

mjschmidt:
Sure. If you delete the operator before the cr you put kubernetes into a stuck state when it goes to delete the namespace. The CR needs to be patched to make its deletion true before the namespace will be cleaned up.

Programmatically, if the operator is deleted and the deployment state is set to zero, it should delete all cdi custom resources before terminating as part of its shut down process. It can't just be done on pod deletion for obvious reasons (:

mjschmidt:
Sure. If CDI Operator Pod is terminated and desired deployment state of the cdi operator in kubernetes is set to zero, then the operator should delete all cdi custom resources and the cdi custom resource definitions.

mjschmidt:
closing in favor of https://github.com/kubevirt/demo/pull/131

PR #288, opened 2019-09-16
Update monitoring documentation to reflect current behavior

stu-gott:
Documents https://github.com/kubevirt/kubevirt/pull/2573

There were a couple of user-facing notes changed, now that KubeVirt automatically creates the ServiceMonitor resource.

Signed-off-by: Stu Gott <sgott@redhat.com>


MarSik:
@stu-gott Does this mean we no longer case about pure k8s cluster? Because there is no ServiceMonitor there, is it?

stu-gott:
@MarSik of course we care. It's always possible to add prometheus to a pure k8s cluster, it's just not there by default.

The distinction this PR makes is that it's no longer the admin's responsibility to create the ServiceMonitor resource, as KubeVirt will take care of that automatically.

PR #291, opened 2019-09-18
Some adjustments to make copy n paste easier

fabiand:
Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>


administration/intro.adoc
```diff
@@ -106,11 +106,10 @@ The installation uses the KubeVirt operator, which manages lifecycle of all Kube
 
 [source,bash]
 ----
-$ export RELEASE=v0.17.0
-# creates KubeVirt operator
-$ kubectl apply -f https://github.com/kubevirt/kubevirt/releases/download/${RELEASE}/kubevirt-operator.yaml
-# creates KubeVirt KV custom resource
-$ kubectl apply -f https://github.com/kubevirt/kubevirt/releases/download/${RELEASE}/kubevirt-cr.yaml
+# Deploy the KubeVirt operator
+$ kubectl apply -f https://github.com/kubevirt/kubevirt/releases/download/v0.21.0/kubevirt-operator.yaml
```
iranzo: Should we make this to take 'latest' release like what we do on minikube or katacoda?

fabiand: Fair point

fabiand: But no, as latest does point to the latest release, which is not the semver highest release
(so latest could point to 0.13.* if we build it after 0.24)

PR #292, opened 2019-09-18
Generate all releases doc

### LINKED ISSUE #259: We don't have docs based on release versions
/kind bug

It is hard to determine if specific features exist in a kubevirt version `X`.

iranzo:
Hopefully....

Closes #259

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 docs: clean
+	scripts/generate_distro_map.sh
 	asciibinder package
 	mv _package/community docs
 
--- a/scripts/generate_distro_map.sh
+++ b/scripts/generate_distro_map.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+# Description: This script generates a distro_map.yml with mentions to each release to build
+
+releases() {
+curl  "https://api.github.com/repos/kubevirt/kubevirt/tags" 2>&1|grep "name"|grep "v[0-9]*\.[0-9]*\.0\""|awk -F ":" '{print $2}'|tr -d '",v'|cut -d "." -f 1-2|grep -v 0.16|grep -v 0.15|grep -v 0.14
+}
+
+gen_distromap() {
+  {
+  # Pregenerate local branches for ascii_binder to work
+  git fetch --all
+  for REL in $(releases);
+  do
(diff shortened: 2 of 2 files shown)


iranzo:
Raised issue to ascii_binder https://github.com/redhataccess/ascii_binder/issues/152

iranzo:
@rmohr do you happen to know where to check that 'older releases' have been uploaded?

The PR rendering is at https://5d834cae267a850008b3bd14--kubevirt-user-guide.netlify.com/latest/welcome/index.html

But the actual documentation process (we haven't touched at all), is at https://kubevirt.io/user-guide/docs/latest/welcome/index.html but I cannot see 'older' releases

Thanks,
Pablo

rmohr:
@iranzo can you explain what you mean by `older` releases?

iranzo:
@rmohr kubevirt is now 0.22, but if we want to provide 'access' to 0.17 docs just in case someone is using 0.17 and wants to check the doc for that release, they can't do that on the website.

One of the points of using current build system (asciibinder) is the provided chance to generate docs for all releases that exist, but in the website only 'latest' is shown.

rmohr:
> One of the points of using current build system (asciibinder) is the provided chance to generate docs for all releases that exist, but in the website only 'latest' is shown.

You would have to correlate dates to reconstruct the documentation as it was at the release time. We did not preserve old docs. Mostly we just wrote something like `since version X you can do Y`. But that obviously does not scale ...

iranzo:
> > One of the points of using current build system (asciibinder) is the provided chance to generate docs for all releases that exist, but in the website only 'latest' is shown.
> 
> You would have to correlate dates to reconstruct the documentation as it was at the release time. >We did not preserve old docs. Mostly we just wrote something like `since version X you can do Y`. >But that obviously does not scale ...

This PR was to setup the documentation process to take care of it, but also requires coordination from the development side to create branches for each documentation 'release' (see the Issue from  https://github.com/kubevirt/user-guide/issues/259 ).

I did send an email to the person in charge of OCP documentation (bumped the thread with you on CC) to see how this should be done for u/d approach

fabiand:
Please sign your commits, then this is a pretty nice contribution!

PR #296, opened 2019-10-21
Fixed KubeVirt CR monitoring example and formatting

slintes:
There is no `AdditionalProperties` on the KubeVirt Spec


slintes:
ping @stu-gott

Issue #298, opened 2019-11-08
API Reference link broken


leslie-wang:
**What happened**:  API Reference at home page seems broken. 

**What you expected to happen**: It points to correct link, and I can open to browse all APIs.

leslie-wang:
I mean this one
![image](https://user-images.githubusercontent.com/2538087/68488732-8ef39000-01fa-11ea-8c8d-a0496c8bf708.png)

PR #299, opened 2019-11-08
Update links to use kubevirt.io without www

codificat:
The alias `www.kubevirt.io` is currently broken (kubevirt/kubevirt.github.io#346), so the links that use it don't work.

This updates 3 links in the user guide to just use `kubevirt.io` instead of `www.kubevirt.io` as a workaround.

/kind bug

Fixes: #298


fabiand:
@codificat please add your isgnature to the commits

`$ git rebase -i` can help

iranzo:
Rebased on top of clean master so that only one commit is there

/lgtm
/approve

PR #305, opened 2020-01-27
Added user-guide for dedicated cpu for QEMU Emulator

vatsalparekh:
Signed-off-by: Vatsal Parekh <vparekh@redhat.com>

This explains the feature to define dedicated CPI for QEMU manager

Explains https://github.com/kubevirt/kubevirt/pull/2870/

### DIFF
--- a/creating-virtual-machines/dedicated-cpu.adoc
+++ b/creating-virtual-machines/dedicated-cpu.adoc
@@ -95,6 +95,31 @@ spec:
 [...]
 ----
 
+Requesting dedicated CPU for QEMU emulator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A number of QEMU threads, such as QEMU main event loop, async I/O operation completion, etc., also execute on the same physical CPUs as the VMI's vCPUs. This may affect the expected latency of a vCPU.
+In order to enhance the real-time support in KubeVirt and provide improved latency, KubeVirt will allocate an additional dedicated CPU, exclusively for the emulator thread, to which it will be pinned. This will effectively "isolate" the emulator thread from the vCPUs of the VMI.
+
+This functionality can be enabled by specifying `isolateEmulatorThread: true` inside VMI spec's `Spec.Domain.CPU` section.
+Naturally, this setting has to be specified in a combination with a `dedicatedCpuPlacement: true`.
+
+Example:
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha3
+kind: VirtualMachineInstance
+spec:
+  domain:
+    cpu:
+      dedicatedCpuPlacement: true
+      isolateEmulatorThread: true
+    resources:
+      limits:
+        cpu: 2
+        memory: 2Gi
+----
+
 Identifying nodes with a running CPU manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


creating-virtual-machines/dedicated-cpu.adoc
```diff
@@ -95,6 +95,31 @@ spec:
 [...]
 ----
 
+Requesting dedicated CPU for QEMU emulator
```
enp0s3: @vatsalparekh Are there any penalties or side effects that may appear while using this feature? I think it worth mentioning it if such exist. Otherwise why not always use it?

vatsalparekh: I would look to @vladikr for this

vladikr: @enp0s3 using this feature the VMI will allocate an additional physical CPU. It's a trade-off that the user needs to be aware of. Definitely not something we should do blindly by default. However, those VMIs which require a more deterministic latency will need to consider moving unnecessary threads aways from the pCPUs that are dedicated for their vCPUs.

vatsalparekh: @enp0s3 @vladikr so any changes suggested here?

creating-virtual-machines/dedicated-cpu.adoc
```diff
@@ -95,6 +95,31 @@ spec:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Apart from VMI's processes, there are additional threads of the QEMU process, such as QEMU main event loop, async I/O operation completion, migration data transfer and more which affect latency and overall performance of the VMIs. In order to enhance the real-time support in KubeVirt, the QEMU emulator threads need to be "isolated" - pinned to a dedicated physical CPU that is not used by any of the VMI virtual CPUs.
+
+This can be defined by specifying `isolateEmulatorThread: true` inside VMI spec's `Spec.Domain.CPU`.
+This will allocate an additional dedicated CPU on the pod exclusively for the emulator thread.
+To use this, `dedicatedCpuPlacement: true` also needs to be there on the VMI spec.
```
vladikr: Thanks, it looks good. However, I think I would write it a bit differently, perhaps something like the following:

A number of QEMU threads, such as QEMU main event loop, async I/O operation completion, etc., also execute on the same physical CPUs as the VMI's vCPUs. This may affect the expected latency of a vCPU.
In order to enhance the real-time support in KubeVirt and provide improved latency, KubeVirt will allocate an additional dedicated CPU, exclusively for the emulator thread, to with it will be pinned. This will effectively "isolate" the emulator thread from the vCPUs of the VMI.

This functionality can be enabled by specifying `isolateEmulatorThread: true` inside VMI spec's `Spec.Domain.CPU` section.
Naturally, this setting has to be specified in a combination with a `dedicatedCpuPlacement: true`.

@stu-gott  what do you think?

stu-gott: "to *which* it will be pinned", other than that I like your suggestion @vladikr

vatsalparekh: @vladikr @stu-gott Updated with your suggestions, back to review

creating-virtual-machines/dedicated-cpu.adoc
enp0s3: @vatsalparekh LGTM

PR #300, opened 2019-12-03
update image upload doc to be more DataVolume centric

mhenriks:
Signed-off-by: Michael Henriksen <mhenriks@redhat.com>

`virtctl image-upload` will now create DataVolumes, so update doc accordingly.

PR #306, opened 2020-01-27
update image upload doc to be more DataVolume centric

aglitke:
`virtctl image-upload` will now create DataVolumes, so update doc accordingly.

This is a rebase of #300 which has been reviewed and approved.

Signed-off-by: Michael Henriksen <mhenriks@redhat.com>


slintes:
see #312

PR #307, opened 2020-02-12
State PVC requirements

iranzo:
Indicate PVC requirements

Signed-off-by: Pablo Iranzo Gómez <Pablo.Iranzo@gmail.com>

### DIFF
--- a/administration/live-migration.adoc
+++ b/administration/live-migration.adoc
@@ -32,6 +32,11 @@ data:
   feature-gates: "DataVolumes,LiveMigration"
 ....
 
+## Limitations
+
+- Virtual machines using a PersistentVolumeClaim (PVC) must have a shared ReadWriteMany (RWX) access mode to be live migrated.
+- Live migration is not allowed with a pod network binding of bridge interface type (https://kubevirt.io/user-guide/docs/latest/creating-virtual-machines/interfaces-and-networks.html)
+
 ## Initiate live migration
 
 Live migration is initiated by posting a VirtualMachineInstanceMigration (VMIM)


administration/live-migration.adoc
```diff
@@ -32,6 +32,8 @@ data:
   feature-gates: "DataVolumes,LiveMigration"
 ....
 
+Additionally, Virtual machines must have a PersistentVolumeClaim (PVC) with a shared ReadWriteMany (RWX) access mode to be live migrated.
```
vladikr: @iranzo Thanks for this clarification. However, I think we should further clarify it as having a PVC is not a requirement for live migration - it can have any of the supported disks. Only if the VM uses a PVC it should be RWX.

iranzo: ok, let me rephrase

administration/live-migration.adoc
```diff
@@ -32,6 +32,8 @@ data:
   feature-gates: "DataVolumes,LiveMigration"
 ....
 
+Additionally, Virtual machines using a PersistentVolumeClaim (PVC) must have a shared ReadWriteMany (RWX) access mode to be live migrated.
```
vladikr: There is a similar comment on line 61, in "Condition and migration method"

administration/live-migration.adoc
vladikr: Thanks :)

administration/live-migration.adoc
iranzo: Should we move that one to the upper part to make it clear that is a requirement and not just a 'rejection' ?

vladikr: Perhaps we need a new section with "Limitations", as we also won't allow a migration with a bridge interface attached to a pod network...[1]

https://kubevirt.io/user-guide/docs/latest/creating-virtual-machines/interfaces-and-networks.html

vladikr: There is also another PR[1] in progress that will add new limitations.

[1] https://github.com/kubevirt/kubevirt/pull/2982

iranzo: I'll add mention to it or should we have that other 'on watch' to update when merged?

vladikr: I'll add a comment in the PR ^^ to update this section after it's merged.

iranzo: Done!

PR #311, opened 2020-02-24
Update instructions to the latest CDI

maya-r:
I've considered making it a command that always picks the latest, but it feels too magical.


maya-r:
Closing as an update also appears in #306.

PR #314, opened 2020-02-26
Added pausing docs

slintes:
Better late than never: added doc of `virtctl pause` and `unpause` command


slintes:
will create a new PR...

/close

PR #318, opened 2020-03-02
Added pausing docs

slintes:
Better late than never: added doc of virtctl pause and unpause command.

Replaced #314 after markdown migration.


slintes:
/assign @davidvossel 

can you have a look please? Would like to get this merged before the next big refactoring ;)

PR #320, opened 2020-03-24
Fix some typos

iranzo:
Fix some typos


rmohr:
tide is not enabled here :D

PR #346, opened 2020-08-06
Minor typo correction & change env var to be consistent with updating doc

asachs01:
A minor typo fix. The main thing here is changing the `VERSION` environment variable to `RELEASE` so that it's consistent with the updating and deleting doc. 

Signed-off-by: Aaron Sachs <aaronm.sachs@gmail.com>


fabiand:
Thanks!

Issue #359, opened 2020-09-24
Migrate to a different markdown engine


mazzystr:
/kind enhancement

### The reasons for this are numerous...
* There is no local tooling for docsifyjs
* Links from kubevirt.io to user-guide get through CI and surprise users with errors
* Contributors have to have knowledge of three different markdowns (GitHub, Jekyll's kramdown, docsifyjs) when working with KubeVirt
* Current documentation hierarchy is not ideal and needs to be addressed
* Cross git repo documentation is not connected to website / user-guide

### A possible alternative is mkdocs
* mkdocs can render the page upon client request.  This is useful when content is being developed.  Contributors can view updates in real time without having to restart the process.  
* I love the readthedocs theme.  This is inline with many projects across the internet.
* The rendering of pages is a single DOM with css tables used to render each panel
* Clients get the ability to obtain proper http return codes when retrieving pages from the server.  Example: if a non-existent page is retrieved the client is given a 404 http error.
* Search works quite well.  Direct pages turn up as first item.  Heading selectors turn up as the first item, the top level page as the second item, all other heading selectors of that page follow in order, then other pages that have the search term mentioned.
* CI can examine page contents very easily with tools like [casperjs|https://casperjs.org]
* Markdown is a similar style as GitHub.  This reduces the number of markdowns contributors need to know.

### CONS
* There are no official mkdocs contanier images.  This means KubeVirt will have to maintain it's own image build workflow.
* The content migration will be significant... 45 pages in user-guide and also 9 links from kubevirt.io.  This is not a unique problem to mkdocs.

ref: Red Hat internal Jira card [KNIECO-2576](https://issues.redhat.com/browse/KNIECO-2576)

mazzystr:
Screenshots of mockup are attached...
<img width="1344" alt="1" src="https://user-images.githubusercontent.com/2479463/94181090-d22b1180-fe53-11ea-9429-7e54dacbdcbd.png">
<img width="1172" alt="2" src="https://user-images.githubusercontent.com/2479463/94181105-d6efc580-fe53-11ea-9dba-86208c11afc1.png">

mazzystr:
Here is the current migration plan....

* [DONE] Create GitHub Issue
https://github.com/kubevirt/user-guide/issues/359
* [DONE] Create proof of concept.  Attach mockup screenshots to issue
* [DONE] Create Dockerfile to run mkdocs image
* * [?] tagging strategy
[...]

PR #361, opened 2020-10-12
Initial commit for OWNERS file

mazzystr:
Initial commit for OWNERS file

note: the tagging is set to the future user of mkdocs which will use directories docs and site.

See user-guide [issues/359](https://github.com/kubevirt/user-guide/issues/359)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,43 @@
+filters:
+  ".*":
+    reviewers:
+      - davidvossel
+      - vladikr
+      - rmohr
+      - stu-gott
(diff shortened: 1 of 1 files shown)


codificat:
Should we split the list of reviewers/approvers by topic? so we can have relevant experts as default reviewers.

That would probably require some restructure of the contents by topic though... so maybe in a separate step/PR?

rmohr:
/lgtm

I think we can start with it like this.

rmohr (approved):

PR #380, opened 2021-01-29
Document watchdog usage

rmohr:
Document the basic usage of watchdogs in kubevirt.

### DIFF
--- a/creation/probes.md
+++ b/creation/probes.md
@@ -1,5 +1,5 @@
-Configure Liveness and Readiness Probes
-=======================================
+Configure Liveness Probes, Readiness Probes and Watchdogs
+=========================================================
 
 It is possible to configure Liveness and Readiness Probes in a similar
 fashion like it is possible to configure [Liveness and Readiness Probes
@@ -15,6 +15,11 @@ Readiness Probes are an indicator for Services and Endpoints if the
 VirtualMachineInstance is ready to receive traffic from Services. If
 Readiness Probes fail, the VirtualMachineInstance will be removed from
 the Endpoints which back services until the probe recovers.
+
+Watchdogs focus on ensuring that an Operating System is still responsive. They
+complement the probes which are more workload centric. Watchdogs require kernel
+support from the guest and additional tooling like the commonly used `watchdog`
+binary.
 
 Define a HTTP Liveness Probe
 ----------------------------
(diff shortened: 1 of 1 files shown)


fabiand:
@mazzystr do you know why the test are failing?

rmohr:
> @mazzystr do you know why the test are failing?

I see 

```
Error: Invalid value for '-f' / '--config-file': Could not open file: mkdocs.yml: No such file or directory
```

@mazzystr maybe that tells you something

codificat:
There are known issues and WIP to fix them for the CI around the user-guide, as part of the migration to a new site renderer engine (mkdocs).

@mazzystr can provide more details, but one thing that I believe should happen here is that the PR should be against the `source` branch instead of `master`.

@rmohr can you point this PR against `source`?

rmohr:
> @mazzystr can provide more details, but one thing that I believe should happen here is that the PR should be against the `source` branch instead of `master`.

If it is due to a switch from `master` to e.g. `main` then I would prefer `main` which seems to be the bigger consensus. If that is not the reason, I am not sure I like this single  exception in the whole org.

codificat:
> > @mazzystr can provide more details, but one thing that I believe should happen here is that the PR should be against the `source` branch instead of `master`.
> 
> If it is due to a switch from `master` to e.g. `main` then I would prefer `main` which seems to be the bigger consensus. If that is not the reason, I am not sure I like this single exception in the whole org.

The plan is to use `source` here in the same way we use it in the website (`kubevirt/kubevirt.github.io` repo). Both repos (user-guide and website) are served through github pages.

In other words: the switch to `source` makes this consistent with the other repo in the org that publishes a static website... makes sense?

OTOH the plan was to render the static site to a new `gh-pages` branch instead of `master` (the website uses `master`). AFAIK this is the default target branch on sites that use `mkdocs`. Obviously, no other repo in the org would be using `mkdocs`... should we use `master` or `main` for the target rendered documentation pages?  TBH it would still be a different type of `master` branch anyway: the default branch in this repo would switch to `source` the same way as in the website repo.

mazzystr:
Please hold off on lgtm and approve on this pull request.  We are going through the final steps of migrating user-guide to the mkdocs markdown engine.   Many updates to CI are being made as we can see here with all the failed checks.  We don't want to have to unwork all that work.

Thanks!

mazzystr:
We had a discussion re the pending content pull_reqs.  Repo owner will manually merge to avoid rework on the CI system.  I will rebase the pull/372.  

Hold label is removed @codificat

PR #396, opened 2021-02-23
host devices: case consistency

dankenigsberg:
I suspect that the host device name in the VM spec must match the one
exposed by KubeVirt device plugin.

Please do not merge before approval by @vladikr.


vladikr (commented):
Oh, yes, thanks Dan!
/lgtm

PR #404, opened 2021-03-05
Fix non-ascii characters

mazzystr:
Fix non-ascii characters

Not sure if anyone noticed but HTMLproofer does not like non-ascii (UTF-8) characters at all.  I tried messing with various aspects of the image and how HTMLproofer operates and there's just no way.  We have to make sure that markdown contains ONLY ascii characters.


docs/virtual_machines/disks_and_volumes.md
```diff
@@ -264,24 +264,24 @@ datasource may look like this:
...
 remain persistent between restarts.
 
-A `PersistentVolume` can be in “filesystem” or “block” mode:
+A `PersistentVolume` can be in "filesystem" or "block" mode:
 
 -   Filesystem: For KubeVirt to be able to consume the disk present on a
-    PersistentVolume’s filesystem, the disk must be named `disk.img` and
+    PersistentVolume's filesystem, the disk must be named `disk.img` and
     be placed in the root path of the filesystem. Currently the disk is
-    also required to be in raw format. &gt; **Important:** The
```
stu-gott, line 276: unsure on this one. I'd assume markdown takes care of that?

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -906,17 +906,17 @@ additional context switching is incurred for each thread.
 
 Disks with `dedicatedIOThread` set to `true` will not use the shared
 thread, but will instead be allocated an exclusive thread. This is
-generally useful if a specific Disk is expected to have heavy I/O
-traffic, e.g. a database spindle.
+generally useful if a specific Disk is expected to have heavy I/O traffic
+ie. a database spindle.
```
stu-gott, line 910: e.g. is technically more correct here. a database spindle isn't the only case, just an example of one.

docs/virtual_machines/presets.md
```diff
@@ -55,8 +57,8 @@ applied to a `VirtualMachineInstance`, the `VirtualMachineInstance` will
 be marked with an Annotation upon completion.
 
 Any domain structure can be listed in the `spec` of a
-`VirtualMachineInstancePreset`, e.g. Clock, Features, Memory, CPU, or
-Devices such as network interfaces. All elements of the `spec` section
+`VirtualMachineInstancePreset` (Clock, Features, Memory, CPU, or
```
stu-gott, line 60: I don't think this list is all encompasing. To me the parenthesis implies a greater exactness to the list of examples--but I don't feel very strongly on this point if we prefer it this way.

docs/virtual_machines/virtual_hardware.md
```diff
@@ -1,7 +1,7 @@
 # Virtual hardware
 
 Fine-tuning different aspects of the hardware which are not device
-related (BIOS, mainboard, …) is sometimes necessary to allow guest
+related (BIOS, mainboard, etc) is sometimes necessary to allow guest
```
stu-gott, line 4: "etc." is always spelled with a period.

stu-gott (commented):
Impressively comprehensive PR. thank you!

A few minor nits.

docs/virtual_machines/disks_and_volumes.md
mazzystr: There was some kind of special character in this line that doesn't show up under normal text editing tools. HTMLProofer would fail with a non-ascii character error.  

The following grep pointed to this line
```
for i in `find docs/ -type f | grep ".md"`; do grep --color='auto' -P -n "[^\x00-\x7F]" $i && echo $i; done
```

I couldn't figure out where the error was.  Replacing the entire line cleared the grep

docs/virtual_machines/disks_and_volumes.md
mazzystr: Reverting the txt passes non-ascii check.  I'll just do that and keep the diff minimal

docs/virtual_machines/presets.md
mazzystr: Reverting the txt passes non-ascii check. I'll just do that and keep the diff minimal

stu-gott (dismissed):
/lgtm

PR #408, opened 2021-03-09
disks: configmap: clarify sentence about update

dankenigsberg:
The replaced text mislead a user, I hope that the suggestd one is
clearer about the bit that is not supported (update of ConfigMap disk
while VMI is running).

Signed-off-by: Dan Kenigsberg <danken@redhat.com>


mazzystr:
/approve
/lgtm

Thank you for your contribution

PR #411, opened 2021-03-10
host-devices.md: fix up mediated device description

jsquyres:
Correct the path names and descriptions of how mediated devices are
matched to corresponding Linux devices based on `mdevNameSelector`.

Signed-off-by: Jeff Squyres <jeff@squyres.com>

@mazzystr In our discussion yesterday, I was just concerned about getting the formatting for this chunk of text correct, and I failed to notice that the text was incorrect about the directory name that it specified.  While I was fixing that, I added a little more explanation about how the matching is done.


mazzystr:
/approve
/lgtm

Thanks for the update

PR #440, opened 2021-05-17
Fix command adding entry to hosts file

yakiduck:
That command may not work expectedly.

### DIFF
--- a/docs/operations/containerized_data_importer.md
+++ b/docs/operations/containerized_data_importer.md
@@ -148,7 +148,7 @@ Use the `openssl` command to view the names of the cdi-uploadproxy service.
 Adding the following entry to the /etc/hosts file, if it provides name resolution, should fix this issue.
 Any service that provides name resolution for the system could be used.
 
-    cat "192.168.39.32  cdi-uploadproxy" >> /etc/hosts
+    echo "192.168.39.32  cdi-uploadproxy" >> /etc/hosts
 
 The upload should now work.
 


docs/operations/containerized_data_importer.md
```diff
@@ -148,7 +148,7 @@ Use the `openssl` command to view the names of the cdi-uploadproxy service.
 Adding the following entry to the /etc/hosts file, if it provides name resolution, should fix this issue.
 Any service that provides name resolution for the system could be used.
 
-    cat "192.168.39.32  cdi-uploadproxy" >> /etc/hosts
+    echo "192.168.39.32  cdi-uploadproxy" >> /etc/hosts
```
mazzystr, line 151: Oh no.  static ips slipped in to this doc.  That will cause confusion also.

Would you mind reviewing this whole section and changing any instance of 192.168.39.32 to <url to upload proxy service> to follow the format of the rest of the doc.  You could variablize that as well to make the cmds even cleaner.

mazzystr (changes requested):

docs/operations/containerized_data_importer.md
yakiduck: Hi. In this section **Does not contain any IP SANs**, the IP address 192.168.39.32 is just for counterexample, so I think it is  no necessary to variablize. Besides, that IP can be pod IP, node IP, ClusterIP etc. I don't know which kind that example could be and variablize.

mazzystr: ok

PR #447, opened 2021-05-26
Document how to enable a feature gate

### LINKED ISSUE #446: Procedure needed to enable a feature gate
There are many mentions of enabling feature gates in user-guide.  Information on how to exactly enable a feature gate is hard to find.

Please create a simple procedure on enabling a feature gate.  This can be a new section in an existing page or a new page.

maiqueb:
Fixes #446


maiqueb:
I have yet to link this from the features that require a feature gate.

/cc @mazzystr

docs/operations/activating_feature_gates.md
```diff
@@ -0,0 +1,34 @@
...
+END
+
+kubectl apply -f enable-feature-gate.yaml
+```
+
+**Note:** the name of the feature gates is case sensitive.
+
+## List of feature gates
+The list of feature gates (which evolve in time) can be checked directly from
+the [source code](https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-config/feature-gates.go#L26).
```
maiqueb, line 33: Is there a way to show a "preview" of the link in markdown ? IMO that would be helpful - I'd rather not have to maintain the list of gated features in this file.

mazzystr: Absolutely.  See README in the main repo

The basic process is ...
1. run `make run`
2. edit markdown
3. web browse to http://0.0.0.0:8000
4. review edit
5. git commit and push
6. create pull req

mazzystr: re FeatureGate list .... @davidvossel is there a better way we can maintain this list?

maiqueb (commented):
I've linked the feature gate information in one place I noticed it was missing: https://deploy-preview-447--kubevirt-user-guide.netlify.app/virtual_machines/interfaces_and_networks/#macvtap

If we're OK linking it like that, I'll proceed into replacing the current calls to feature gate activation I've seen so far:
  - https://kubevirt.io/user-guide/operations/live_migration/#enabling-the-live-migration-support
  - https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/#enabling-datavolume-support
  - https://kubevirt.io/user-guide/virtual_machines/virtual_hardware/#enabling-cpu-compatibility-enforcement
  - https://kubevirt.io/user-guide/virtual_machines/dedicated_cpu_resources/#enabling-the-cpu-manager-automatic-identification-feature-gate

docs/operations/activating_feature_gates.md
mazzystr: I completely misunderstood.  Sorry about that.

Try out the info from this post.... https://stackoverflow.com/questions/41966638/jekyll-on-github-pages-include-markdown-in-another-markdown-file

maiqueb: My bad; shouldn't have referred the word "preview". That was awfully misleading  :)

> Try out the info from this post.... https://stackoverflow.com/questions/41966638/jekyll-on-github-pages-include-markdown-in-another-markdown-file

I've looked at the link you've pasted, and afaiu, this would only apply when including a file relative to the one doing the including. Which, I guess means it must be included in the same repo.

It seems that we either link to the source code, or just manually keep the list in sync ... 

Is there any established precedence here ? What are we doing elsewhere in the docs ? I am OK with either option, but would prefer to not have to keep the list synced manually.

maiqueb:
@mazzystr I think this is ready for a review.

Regarding the question in https://github.com/kubevirt/user-guide/pull/447#discussion_r642354341; I think that since we reference to that page mentioning the correct string, we are in a good state. 

It might even be a good option to drop the `## List of feature gates`  section I've introduced.

docs/operations/activating_feature_gates.md
mazzystr: I like how you have found the other references to feature gate and linked to this page.  Let's call the link to the code good enough and move on.  Maybe later we can think of a way to import that list.

Issue #348, opened 2020-08-11
Fix KubeVirt architecture section


mazzystr:
/kind bug

The current [architecture](https://kubevirt.io/user-guide/architecture) page in user-guide does not actually describe the KubeVirt architecture.  Please rebuild this doc. 

Update: Architecture doc is [here|https://github.com/kubevirt/kubevirt/blob/master/docs/architecture.md].  This story just becomes a matter of fixing the web page link which is significantly easier.

ansijain:
@mazzystr I have been looking into this.
Do I need to add external link [here](https://github.com/kubevirt/user-guide/blob/master/docs/.pages) ?
or should I add https://github.com/kubevirt/kubevirt/blob/master/docs/architecture.md in user-guide repo?
Or any other way please guide

PR #449, opened 2021-05-31
Fix Architecture Section

ansijain:
Fixes https://github.com/kubevirt/user-guide/issues/348
Signed-off-by: ansijain <ansi.jain@india.nec.com>


mazzystr:
re https://github.com/kubevirt/kubevirt/pull/5650 ... @fabiand @davidvossel @rthallisey Think we should incorporate this image into the architecture page?

ansijain:
Do you think we should add  [Kubevirt Components Description  ](https://github.com/kubevirt/kubevirt/blob/master/docs/components.md) here?

mazzystr:
> Do you think we should add [Kubevirt Components Description ](https://github.com/kubevirt/kubevirt/blob/master/docs/components.md) here?

@davidvossel I think we should.  How about you?  Something like this should be at the top of the stack so a potential user / architect doesn't have to go looking through the haystack for it.

docs/architecture.md
```diff
@@ -1,5 +1,136 @@
...
+    ~~+---------------------+~~
+      | Orchestration (K8s) |
+      +---------------------+
+      | Scheduling (K8s)    |
+      +---------------------+
+      | Container Runtime   |
+    ~~+---------------------+~~
+      | Operating System    |
+      +---------------------+
+      | (Virtual)           |
```
rthallisey, line 20: kvm

rthallisey (approved):
lgtm

docs/architecture.md
ansijain: Updated KVM in stack

mazzystr:
@ansijain Thank you for your contribution

Issue #454, opened 2021-06-17
Remove all links to Red Hat OpenShift


mazzystr:
I thought we did this already but sure enough I just found another one...

https://kubevirt.io/user-guide/virtual_machines/windows_virtio_drivers/ -> Guest Virtual Machine Device Configuration -> https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/chap-guest_virtual_machine_device_configuration

PR #462, opened 2021-07-13
Updated Links

ansijain:
Fixes https://github.com/kubevirt/user-guide/issues/454
Signed-off-by: ansijain <ansi.jain@india.nec.com>

### DIFF
--- a/docs/virtual_machines/windows_virtio_drivers.md
+++ b/docs/virtual_machines/windows_virtio_drivers.md
@@ -7,6 +7,8 @@ Microsoft Windows running in a fully virtualized guest.
 
 Yes. Without the virtio drivers, you cannot use paravirtualized hardware properly. It would either not work, or will have a severe performance penalty.
 
+For more information about VirtIO and paravirtualization, see [VirtIO and paravirtualization](https://wiki.libvirt.org/page/Virtio)
+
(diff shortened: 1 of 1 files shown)


docs/virtual_machines/windows_virtio_drivers.md
```diff
@@ -5,14 +5,10 @@ Microsoft Windows running in a fully virtualized guest.
 
 ## Do I need virtio drivers?
 
-Yes. Without the virtio drivers, you cannot use
-[paravirtualized](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/virtualization_getting_started_guide/sec-virtualization_getting_started-products-virtualized-hardware-devices#sec-Virtualization_Getting_Started-Products-paravirtdevices)
```
jean-edouard, line 9: Having a link to some kind of definition for the term "paravirtualized" seems useful, some readers may not know what that means.
If that page is not good, maybe we can find something else?

ansijain: @jean-edouard Yes I agree. Do you think we should keep the same link?
I tried finding another link having good information but didn't found any...
Do you have any references ?

cwilkers: [VirtIO](https://wiki.libvirt.org/page/Virtio)'s main landing page goes into the benefits of paravirtualized devices. It does not explicitly define the term paravirtualization, but describes the benefits well. Perhaps this link could be added under the paragraph as a "for more information about VirtIO and paravirtualization, see:"

There is also a [Wikipedia Article](https://en.wikipedia.org/wiki/Paravirtualization) on the concept, but it seems to cover history more than current usage.

ansijain: I have added https://wiki.libvirt.org/page/Virtio link

cwilkers:
Unfortunately, you have removed one vendor link and added another, which goes against the request in #454.
What this exposes, however, is not a bias in the document writer here, but a need for additional supporting documentation at this site so we can replace this link with something local.

I will add my lgtm and open a related issue requesting KubeVirt documentation around VirtIO drivers.

/lgtm

Issue #459, opened 2021-07-09
Document how to use our customization patch interface on the KubeVirt CR in kubevirt


rmohr:
https://github.com/kubevirt/kubevirt/pull/3612 is not documented. Or at least I can't find anything in the user-guide.

rmohr:
/assign @ashleyschuett 

could you have a look?

PR #467, opened 2021-07-20
add user guide for patch and flag customization options

ashleyschuett:
Added documentation on the different customization options exposed in kubevirt. 

Fixes https://github.com/kubevirt/user-guide/issues/459

### DIFF
--- a/docs/operations/customize_components.md
+++ b/docs/operations/customize_components.md
@@ -1,0 +1,82 @@
+
+## Customize KubeVirt Components
+
+
+### Customize components using patches
+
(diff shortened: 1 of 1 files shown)


dhiller (approved):
Thanks for the documentation, looks very good in general! 

One nit: At the end of the customize flags section I would have expected another explanation of what either the flags do or a reference to the source somehow. Not sure if that is really helpful, just a thought.

ashleyschuett:
@dhiller thanks for the review and suggestion. I have updated the flags section to show and example and the out put produced. Let me know if it makes sense or could be made more intuitive :)

dhiller (approved):
Thanks for the updates, looks good to me!

/approve

PR #471, opened 2021-08-09
Fix: fix path in architecture.md from absolute path to related

tjjh89017:
Image path is incorrect with starting slash
follow other page to fix it and remove the starting slash to be related path and be migratable again.

Signed-off-by: Date Huang <tjjh89017@hotmail.com>

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -67,11 +67,11 @@ users to speak to Kubernetes, but modify VMIs.
 The following diagram illustrates how the additional controllers and daemons
 communicate with Kubernetes and where the additional types are stored:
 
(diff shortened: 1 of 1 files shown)


cydu-cloud:
> @mazzystr
> thanks a lot
> i rebase it to master branch
> 
> it seems pic's issue is fixed in #474
> but they changed it into abs path
> i think it's better to use related path because it can be still migrateable
> please check it

Sorry, didn't see your PR before. I created #479 to fix that with ralative path.

mazzystr:
Ah ok.  I'll close then.

.....And welcome to KubeVirt.  We hope you'll stay awhile.  :)

tjjh89017:
ok tks guys!

PR #498, opened 2021-10-28
Document NVMe passthrough

alicefr:
Add documentation for NVMe passthrough and its limitations for KubeVirt

### DIFF
--- a/docs/virtual_machines/host-devices.md
+++ b/docs/virtual_machines/host-devices.md
@@ -115,3 +115,60 @@ spec:
       - deviceName: intel.com/qat
         name: quickaccess1
 ```
+
+## NVMe PCI passthrough
+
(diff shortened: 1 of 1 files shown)


mazzystr:
This section looks and I'm happy to tag.

Question before doing so tho.  What do you think about the ordering of the page as whole top to bottom?  Do you think we should refactor so that it feels less zig-zag?  For instance I see there is a "Starting Virtual machine" section stuck right ion the middle then we get back to configuring for NVMe.

alicefr:
@mazzystr yes I see your point! The only thing is that I have an example of VMI with the NVMe and moving the section before `Starting a Virtual Machine` sounds a bit weird to me. Either I remove the VMI example or... I'm not sure :)

dhiller (approved):
Thanks!

vladikr:
/lgtm
Thanks @alicefr !

alicefr:
@mazzystr any further comments on this?

mazzystr:
All good with me!  Thanks!

/lgtm
/approve

PR #501, opened 2021-11-03
Typo: Slack vs Stack Overflow

jean-edouard:
Seems weird to participate on Slack Overflow, I assume we meant Slack?


mazzystr:
Yup, definitely a mistake.  We aren't supporting Stack Overflow due to manpower.  Thanks for catching it.

/lgtm
/approve

PR #509, opened 2022-01-03
add limitations when memfd is supported

alicefr:
A couple of users asked about the support for memfd because they got the error `unsupported configuration: hugepages is not supported with memfd memory source`. This limitation is due to a too old host kernel. Adding this information could be useful to prevent other doubts.

### DIFF
--- a/docs/virtual_machines/virtual_hardware.md
+++ b/docs/virtual_machines/virtual_hardware.md
@@ -605,6 +605,8 @@ regular memory it will use node hugepages of the size of `2Mi`.
 
 -   requested memory must be divisible by hugepages size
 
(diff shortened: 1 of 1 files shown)


jordigilh:
@vladikr @fabiand @rmohr can you take a look at this change for the user's guide? This problem was found recently by someone who was attempting to deploy a VM with the real time knob enabled and failed because their kernel was below version 4.14 (CentOS 7.9).
Adding this to the user's guide should help others who might face the same issues in the future.

Thanks :smile:

vladikr:
Look good to me, thank you @alicefr 
/approve

PR #514, opened 2022-01-28
Correct the spacing for device tagging snippet

### LINKED ISSUE #512: Device Role Tagging VM spec snippet is wrong
The VM spec on the example of [[1]](https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#device-role-tagging)
is wrong.
The `tag` should be part of each device object instead on the `domain` level.
It seems like interpretation problem of the markdown file since it looks right on the file itself [[2]](https://github.com/kubevirt/user-guide/blob/master/docs/virtual_machines/startup_scripts.md#device-role-tagging)

jbpratt:
Signed-off-by: jbpratt <jbpratt78@gmail.com>

### DIFF
--- a/docs/virtual_machines/startup_scripts.md
+++ b/docs/virtual_machines/startup_scripts.md
@@ -599,7 +599,7 @@ spec:
         name: default
       - bridge: {}
         name: ptp
-	tag: ptp
+        tag: ptp
       - name: sriov-net
         sriov: {}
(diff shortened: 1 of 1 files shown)


jbpratt:
closes #512

cwilkers:
Thanks for the PR! Sorry about the idle time

PR #528, opened 2022-04-06
docs, network: document bridge interface invalid CNIs

maiqueb:
There are recurring issues being created on KubeVirt related to some CNIs being used which we know do not work with our bridge binding option - the most common choice for secondary interfaces. 

Example issues: [[0](https://github.com/kubevirt/kubevirt/issues/5483)], [[1](https://github.com/kubevirt/kubevirt/issues/7001)], but there are more ...

For instance, we know that neither macvlan nor ipvlan plugins will work when the plugin configuration features IPAM for secondary bridge interfaces.

This commit adds a section to the user guide describing which plugins are known not work, along with a brief explanation of why.

### DIFF
--- a/docs/virtual_machines/interfaces_and_networks.md
+++ b/docs/virtual_machines/interfaces_and_networks.md
@@ -181,6 +181,23 @@ spec:
       networkName: macvlan-test
 ```
 
+#### Invalid CNIs for secondary networks
+The following list of CNIs is known **not** to work for bridge interfaces -
+which are most common for secondary interfaces.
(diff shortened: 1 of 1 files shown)


phoracek:
@maiqueb I would like this section of docs to be the only thing we have to share with users when they hit this issue. While it explains the limitation, it seems too open ended (?). Would it make sense to open a "known issue" tracker on KubeVirt describing in short the problem, and linking this issue from the user-guide? Then people can navigate to kubevirt/kubevirt, discuss it, find linked attempts to fix it, volunteer to resolve the problem.

PR #529, opened 2022-04-07
[Kernel boot]: fix doc - kernel args without image is disallowed

iholder-redhat:
Also, see this PR for implementation: https://github.com/kubevirt/kubevirt/pull/7530


mazzystr:
/approve
/lgtm

Thanks for the update

PR #530, opened 2022-04-10
Bump docs.openshift links to 4.10

kaovilai:


cwilkers (approved):
Thanks for the PR/updates!

It could be a TODO for a future update to re-evaluate how often we refer to OpenShift documentation, as KubeVirt should, as much as possible, refer to its own documentation for concepts. It makes sense to point to OCP for OCP specific features, but the virtio drivers in Windows VMs guide, in particular, should be handled in KubeVirt's own docs.

/lgtm
/approve

PR #533, opened 2022-04-14
Add documentation for defining verbosity levels

iholder-redhat:


docs/operations/debug.md
```diff
@@ -0,0 +1,59 @@
...
+In code, the log entry looks similar to: `log.Log.V(verbosity).Infof("...")` while `verbosity` is the minimum verbosity
+level for this entry.
+
+For example, if the log verbosity for some log entry is `3`, then the log would be exposed only if the log verbosity
+is defined to be equal or greater than `3`, or else it would be filtered out.
+
+Currently, log verbosity can be defined per-component or per-node. The most updated API is detailed [here](https://kubevirt.io/api-reference/master/definitions.html#_v1_logverbosity).
+
+### Setting verbosity per Kubevirt component
+One way of raising log verbosity is to manually determine it for the different components in `KubevirtCR`:
```
cwilkers, line 21: I haven't seen this way of referring to the KubeVirt CR, is this in use elsewhere? (If not, it might be better to break it out as the `KubeVirt` `CustomResource` or `KubeVirt` CR)

docs/operations/debug.md
```diff
@@ -0,0 +1,59 @@
+# Debug
+This page contains instructions on how to debug Kubevirt.
```
cwilkers, line 2: For consistency, could you do a search/replace for `s/Kubevirt/KubeVirt/g` ?

docs/operations/debug.md
```diff
@@ -0,0 +1,59 @@
...
+
+This is useful to both Kubevirt developers and advanced users that would like to gain deep understanding on what's
+happening behind the scenes.
+
+## Log Verbosity
```
cwilkers, line 7: In this section, it would be nice to have a guideline (I assume there is guidance in code for devs creating log entries) explaining the full range of debug numbers and what conceptual level of debugging output they might correspond to, e.g. the Unix/Linux syslog of Crit, Error, Warn, Info, Debug. Also, is 6 the maximum used in code? What if somebody sets it to 9 (or 99)?

cwilkers (changes requested):
This looks really useful, I have some minor (take or leave) suggestions and one naming consistency request.

Also, for your consideration, it could complete the document to include an example of checking the logs for one of the components. If that seems too out of scope, perhaps a link to the Kubernetes documentation on reading Pod logs.

docs/operations/debug.md
iholder-redhat: Done, thanks!

docs/operations/debug.md
iholder-redhat: Done, thanks!

iholder-redhat:
>Also, for your consideration, it could complete the document to include an example of checking the logs for one of the components. If that seems too out of scope, perhaps a link to the Kubernetes documentation on reading Pod logs.

Done, thanks for awesome suggestion!

docs/operations/debug.md
iholder-redhat: I totally agree in principle, but in practice there is no clear definition of the exact meaning of the values or the min/max number that can be used. It is clear that higher value means more verbosity, but that's that. I agree we should properly define it in the future.

cwilkers: That's surprising, but yes, it would seem out of scope for this PR then.

PR #539, opened 2022-05-03
Document basic SEV support

vasiliy-ul:
Added documentation on how to run a VM with SEV memory encryption.

### DIFF
--- a/docs/virtual_machines/confidential_computing.md
+++ b/docs/virtual_machines/confidential_computing.md
@@ -1,0 +1,67 @@
+# Confidential computing
+
+## AMD Secure Encrypted Virtualization (SEV)
+
+**FEATURE STATE:** KubeVirt v0.49.0 (experimental support)
+
+[Secure Encrypted Virtualization (SEV)](https://developer.amd.com/sev/) is a feature of AMD's EPYC CPUs that allows the memory of a virtual machine to be encrypted on the fly.
+
+KubeVirt supports running confidential VMs on AMD EPYC hardware with SEV feature.
+
+### Preconditions
+
+In order to run an SEV guest the following condition must be met:
+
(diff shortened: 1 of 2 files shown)


docs/virtual_machines/confidential_computing.md
```diff
@@ -0,0 +1,66 @@
...
+        #cloud-config
+        password: fedora
+        chpasswd: { expire: False }
+    name: cloudinitdisk
+```
+
+### Current limitations
+
+- SEV-encrypted VMs cannot contain directly-accessible host devices (that is, PCI passthrough)
+- Live Migration is not supported
```
alicefr, line 66: We should also specify that the VMs are not attested to avoid any confusion

vasiliy-ul: Good  point. Mentioned it as a limitation.

docs/virtual_machines/confidential_computing.md
```diff
@@ -0,0 +1,67 @@
...
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  labels:
+    special: vmi-fedora
+  name: vmi-fedora
+spec:
+  domain:
+    launchSecurity:
+      sev: {}
```
maiqueb, line 33: are there any interesting parameters to be set here ? Would it make sense to elaborate a bit on it ?

vasiliy-ul: ATM there are no additional parameters. Some will be added for SEV-{ES,SNP} and when attestation is enabled.

maiqueb (approved):
Thank you.

jean-edouard:
/approve
Thank you!

PR #542, opened 2022-05-24
Add more information about Windows answer files

0xFelix:
Links to Window SIM and the answer file references were added to the Sysprep section.


cwilkers (approved):
The links look good to me, thanks for the update!

/lgtm
/approve

PR #544, opened 2022-06-06
Add documentation for memory dump command

ShellyKa13:
Signed-off-by: Shelly Kagan <skagan@redhat.com>

### DIFF
--- a/docs/operations/memory_dump.md
+++ b/docs/operations/memory_dump.md
@@ -1,0 +1,79 @@
+# Virtual machine memory dump  
+
+Kubevirt now supports getting a VM memory dump for analysis purposes.
+
+## Prerequisites
+
+### Hot plug Feature Gate
+
+The memory dump process mounts a PVC to the virt-launcher in order to get the output in that PVC, hence the hot plug volumes feature gate must be enabled. The
+[feature gates](./activating_feature_gates.md#how-to-activate-a-feature-gate)
+field in the KubeVirt CR must be expanded by adding the `HotplugVolumes` to it.
+
+### A PVC to hold the memory dump output
+
+Currently in order to get the memory dump output a pre-existing PVC is required. The size of the PVC must be big enough to hold the memory dump (The VM memory size + 100Mi overhead).
+
+Example for such PVC:
+
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: my-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 2Gi
+  storageClassName: rook-ceph-block
+```
+
+The PVC must be a FileSystem volume mode PVC.
+
+## Virtctl support
+
+### Get memory dump
+
+Now lets assume we have a running VM and the name of the VM is 'my-vm'.
+We can get a memory dump of this VM to the above PVC by using the 'memory-dump get' command available with virtctl
+
+```bash
+$ virtctl memory-dump get my-vm --claim-name=my-pvc
+```
+
+This will dump the memory of the running VM to the given PVC.
+The information of the memory dump will be updated on the VM status.
+```yaml
+    memoryDumpRequest:
+      claimName: memory-dump
+      phase: Completed
+      startTimestamp: "2022-03-29T11:00:04Z"
+      endTimestamp: "2022-03-29T11:00:09Z"
+      fileName: my-vm-my-pvc-20220329-110004
+```
+
+During the process the volumeStatus on the VMI will be updated with the process information such as the attachment pod information and messages, if all goes well once the process is completed, the PVC is unmounted from the virt-launcher pod and the volumeStatus is deleted.
+
+Getting a new memory dump to the same PVC is possible without the need to use the claim-name flag.
+Each memory-dump command will delete the previous dump in that PVC.
+In order to get a memory dump to a different PVC you need to 'remove' the current memory-dump PVC and then do a new get with the new PVC name.
+
+### Remove memory dump
+
+As mentioned in order to remove the associated memory dump PVC you need to run a 'memory-dump remove' command. This will allow you to replace the current PVC and get the memory dump to a new one.
+
+```bash
+$ virtctl memory-dump remove my-vm
+```
+
+## Handle the memory dump
+Once the memory dump process is completed the PVC will hold the output.
+There are several options to get the dump out of the PVC:
+- Copy it from the PVC to your local environment by using virtctl scp command(for more info look for virtctl scp help). This is useful if the VM memory is pretty small.
+- Use Export mechanism by:
+    - Exporting only the PVC (not supported yet)
+    - Include the memory dump in the VMSnapshot and export the whole VMSnapshot (will include both the memory dump and the disks) (not supported yet) 
+
+The output of the memory dump can be used for memory analysis with different tools for example [Volatility3](https://github.com/volatilityfoundation/volatility3)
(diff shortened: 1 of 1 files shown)


docs/operations/memory_dump.md
```diff
@@ -0,0 +1,79 @@
...
+
+Now lets assume we have a running VM and the name of the VM is 'my-vm'.
+We can get a memory dump of this VM to the above PVC by using the 'memory-dump get' command available with virtctl
+
+```bash
+$ virtctl memory-dump get my-vm --claim-name=my-pvc
+```
+
+This will dump the memory of the running VM to the given PVC.
+The information of the memory dump will be updated on the VM status.
```
akalenyu, line 47:
```suggestion
Information regarding the memory dump process will be available on the VM's status section
```

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,79 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: my-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 2Gi
+  storageClassName: rook-ceph-block
```
akalenyu, line 30: Maybe explicitly include volumeMode: Filesystem here too?

akalenyu (commented):
Nice and straightforward

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,79 @@
...
+As mentioned in order to remove the associated memory dump PVC you need to run a 'memory-dump remove' command. This will allow you to replace the current PVC and get the memory dump to a new one.
+
+```bash
+$ virtctl memory-dump remove my-vm
+```
+
+## Handle the memory dump
+Once the memory dump process is completed the PVC will hold the output.
+There are several options to get the dump out of the PVC:
+- Copy it from the PVC to your local environment by using virtctl scp command(for more info look for virtctl scp help). This is useful if the VM memory is pretty small.
```
mhenriks, line 74: Pardon my ignorance, but how does this work?  How is the memory dump accessible from inside the guest?  Does the PVC have to be mounted with virtiofs?

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,79 @@
...
+      claimName: memory-dump
+      phase: Completed
+      startTimestamp: "2022-03-29T11:00:04Z"
+      endTimestamp: "2022-03-29T11:00:09Z"
+      fileName: my-vm-my-pvc-20220329-110004
+```
+
+During the process the volumeStatus on the VMI will be updated with the process information such as the attachment pod information and messages, if all goes well once the process is completed, the PVC is unmounted from the virt-launcher pod and the volumeStatus is deleted.
+
+Getting a new memory dump to the same PVC is possible without the need to use the claim-name flag.
```
awels, line 59: What happens in the case I have not yet used the `--claim-name` flag for the first time and I omit the flag? Will the command fail?

docs/operations/memory_dump.md
brybacki: And I suppose one can also unplug the pvc and use it as additional volume to create the pod with tools needed for troubleshooting.

ShellyKa13: @mhenriks You are right I didnt really explained well here, and I forgot that I used a consumer pod for that (at first when the command didnt unmounted itself I did use copy from the virt-launcher PVC), I should explain better that one can create a consumer pod and then copy the output. and maybe scp is not the right command I just used exec with 'cat'
@brybacki that also a valid option for sure, although is not getting the output out of the PVC, I'll rephrase that so I can add that option too. thanks

docs/operations/memory_dump.md
ShellyKa13: yes, for the first time one must add the claim-name

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,100 @@
...
+        - name: my-pvc
+          persistentVolumeClaim:
+            claimName: my-pvc
+      containers:
+      - name: pod-container
+        image: busybox
+        command: ['/bin/sh', '-c', 'while true; do echo hello; sleep 2;done']
+        volumeMounts:
+          - mountPath: /dev/pvc
+            name: fs-pvc
```
mhenriks, line 93: should be `my-pvc`

ShellyKa13: done

peterclauterbach (changes requested):

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,100 @@
...
+        image: busybox
+        command: ['/bin/sh', '-c', 'while true; do echo hello; sleep 2;done']
+        volumeMounts:
+          - mountPath: /dev/pvc
+            name: my-pvc
+    ```
+- Create a pod with troubleshooting tools that will mount the PVC and inspect it within the pod.
+- Use Export mechanism by:
+    - Exporting only the PVC (not supported yet)
+    - Include the memory dump in the VMSnapshot and export the whole VMSnapshot (will include both the memory dump and the disks) (not supported yet)
```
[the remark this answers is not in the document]
ShellyKa13: actually I dont know whats the best way to mention/track the expansion which should be available pretty soon. @mhenriks @awels you know how I should do that?

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,100 @@
+# Virtual machine memory dump  
+
+Kubevirt now supports getting a VM memory dump for analysis purposes.
```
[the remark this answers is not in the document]
ShellyKa13: @peterclauterbach add a new commit for this two. Is what I added enough?

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,100 @@
...
+As mentioned in order to remove the associated memory dump PVC you need to run a 'memory-dump remove' command. This will allow you to replace the current PVC and get the memory dump to a new one.
+
+```bash
+$ virtctl memory-dump remove my-vm
+```
+
+## Handle the memory dump
+Once the memory dump process is completed the PVC will hold the output.
+You can manage the dump in one of the following ways:
+- Create a consumer pod for the PVC that will mount the PVC as a device, then copy the content out from the pod using `kubectl cp`
```
mhenriks, line 78: I think adding example `kubectl cp` invocation would be helpful

ShellyKa13: done

docs/operations/memory_dump.md
```diff
@@ -0,0 +1,108 @@
...
+        image: busybox
+        command: ['/bin/sh', '-c', 'while true; do echo hello; sleep 2;done']
+        volumeMounts:
+          - mountPath: /dev/pvc
+            name: my-pvc
+    ```
+
+    kubectl cp example:
+    ```bash
+    $ kubectl cp default/consumer-pod:/dev/pvc/my-vm-my-pvc-20220613-144318.memory.dump my-vm-my-pvc-20220613-144318.memory.dump
```
mhenriks, line 101: given the dynamic naming scheme, maybe it makes most sense to simply copy the root dir `/dev/pvc`?

docs/operations/memory_dump.md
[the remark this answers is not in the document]
peterclauterbach: lgtm

docs/operations/memory_dump.md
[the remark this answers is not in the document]
peterclauterbach: no worries, if we don't have a proper future artifact to point to, this text is fine.

PR #401, opened 2021-03-04
Update script to allow version args and change how markdown is popula…

mazzystr:
* Update script to allow version args.  This function is probably a one off.  In the future I expect `git tag | head -1` to work as written.
* Update script to prepend to the file instead of repopulate the file.  There are many spelling mistakes in the git log.  I fixed them in the markdown and updated yaspeller dictionary file where necessary.  I cannot rewrite git log history and I don't want to recreate the markdown file with all the spelling mistakes.  Therefore we prepend and deal with spelling mistakes as they come in.  I don't mind handling this manually until we can automate in the near future.
* Update to markdown to create missing versions.

Fixes #342

PR #546, opened 2022-06-14
Updating latest_release_notes from 0.38.1 to 0.54.0

aburdenthehand:
Batch updating the latest release notes (from 0.38.1 to 0.54.0) using modified `update_changelog` script.

This is following on from #401, I suppose. Looks like this script is meant to be run on a monthly basis (or match release cadence). Should it be part of the release process?

Signed-off-by: aburdenthehand <aburden@redhat.com>

### DIFF
--- a/docs/latest_release_notes.md
+++ b/docs/latest_release_notes.md
@@ -1,5 +1,608 @@
 # Latest release notes
 
+## v0.54.0
+
+Released on: Wed Jun 8 14:15:43 2022 +0000
+
+- [PR #7757][orenc1] new alert for excessive number of VMI migrations in a period of time.
+- [PR #7517][ShellyKa13] Add virtctl Memory Dump command
+- [PR #7801][VirrageS] Empty (`nil` values) of `Address` and `Driver` fields in XML will be omitted.
+- [PR #7475][raspbeep] Adds the reason of a live-migration failure to a recorded event in case EvictionStrategy is set but live-migration is blocked due to its limitations.
+- [PR #7739][fossedihelm] Allow `virtualmachines/migrate` subresource to admin/edit users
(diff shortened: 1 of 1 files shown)


phoracek (changes requested):
It is not really "latest" release notes if we keep last year of versions there.

I would suggest we replace it with a link to https://github.com/kubevirt/kubevirt/releases, so we don't have to update this every month

aburdenthehand:
> It is not really "latest" release notes if we keep last year of versions there.
> 
> I would suggest we replace it with a link to https://github.com/kubevirt/kubevirt/releases, so we don't have to update this every month

That was my first thought as well. However, the benefit of ~mirroring the release notes in the user guide is that it is included in the search bar, and can be easily navigated using the side bar. The github [Find a release](https://github.com/kubevirt/kubevirt/releases) search is significantly poorer imho.

phoracek:
I see. Then I'd just nag to keep only the release notes for recent versions only. Or change the title of the page.

maiqueb:
> > It is not really "latest" release notes if we keep last year of versions there.
> > I would suggest we replace it with a link to https://github.com/kubevirt/kubevirt/releases, so we don't have to update this every month
> 
> That was my first thought as well. However, the benefit of ~mirroring the release notes in the user guide is that it is included in the search bar, and can be easily navigated using the side bar. The github [Find a release](https://github.com/kubevirt/kubevirt/releases) search is significantly poorer imho.

I don't know if it's poorer ... If you want release v0.42.0 you just have to input `"v0\.42\.0"` . That gives the `v0.42.0` release. 
Forces you to use quotes and escape the dots but ... It's easy.

maiqueb:
Digesting both https://github.com/kubevirt/user-guide/pull/546#issuecomment-1156532157 and https://github.com/kubevirt/user-guide/pull/546#issuecomment-1156612607, my vote would go to feature the full release notes, but to change the name (and heading...) of the page.

aburdenthehand:
Created #547 to detach changing the name and updating the related files from updating the missing release notes.

PR #549, opened 2022-06-21
VirtualMachineExport documentation

awels:
Documentation for the VirtualMachineExport feature including PVC/VMSnapshot/VM export and matching virtctl commands.

Signed-off-by: Alexander Wels <awels@redhat.com>

### DIFF
--- a/docs/operations/export_api.md
+++ b/docs/operations/export_api.md
@@ -11,11 +11,64 @@ field in the KubeVirt CR must be expanded by adding the `VMExport` to it.
 
 ### Export token
 
-In order to securely export a Virtual Machine Disk, you must create a token that is used to authorize users accessing the export endpoint. This token must be in a secret that is stored in the namespace the Virtual Machine resides in. The contents of the secret can be passed as a token header or parameter to the export URL. The name of the header or argument is `x-kubevirt-export-token` with a value that matches the content of the secret. The secret can be named any valid secret in the namespace. We recommend you use best security practices when generating this token. 
+In order to securely export a Virtual Machine Disk, you must create a token that is used to authorize users accessing the export endpoint. This token must be in a secret that is stored in the namespace the Virtual Machine resides in. The contents of the secret can be passed as a token header or parameter to the export URL. The name of the header or argument is `x-kubevirt-export-token` with a value that matches the content of the secret. The secret can be named any valid secret in the namespace. We recommend you generate an alpha numeric token of at least 12 characters. The data key should be `token`. For example:
 
-### Export Virtual Machine disk
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: example-token
+data:
+  token: MTIzNDU2Nzg5MGFi
+```
+### Export Virtual Machine volumes
+After you have created the token you can now create a VMExport CR that identifies the Virtual Machine you want to export. You can create a VMExport that looks like this:
 
-After you have created the token you can now create a VMExport CR that identifies the disk of the VM you want to export. In KubeVirt in the end all persisted Virtual Machine disks are stored in a Persistent Volume Claim (PVC) in the namespace the Virtual Machine resides in. Once you have identified the PVC you can create a VMExport that looks like this:
+```yaml
+apiVersion: export.kubevirt.io/v1alpha1
+kind: VirtualMachineExport
+metadata:
+  name: example-export
+spec:
+  tokenSecretRef: example-token
+  source:
+    apiGroup: "kubevirt.io"
+    kind: VirtualMachine
+    name: example-vm
+```
+
+The following volumes present in the VM will be exported:
+
+* PersistentVolumeClaims
+* DataVolumes
+* MemoryDump
+
+All other volume types are not exported. The export will only be available when the VM is not running, this is to prevent corrupt data from being exported. If you want to export data from a running VM, the VM will have to use storage that can be snapshotted, and you can create a Virtual Machine Snapshot and export that while the VM stays running. If an export is active and the VM is started the export will detect this state and terminate the export server.
+
+If the VM contains multiple volumes that can be exported, each volume will get its own URL links. If the VM contains no volumes that can be exported, the VMExport will go into a `Skipped` phase, and no export server is started.
+
+### Export Virtual Machine Snapshot volumes
+
+After you have created the token you can now create a VMExport CR that identifies the Virtual Machine Snapshot you want to export. You can create a VMExport that looks like this:
+
+```yaml
+apiVersion: export.kubevirt.io/v1alpha1
+kind: VirtualMachineExport
+metadata:
+  name: example-export
+spec:
+  tokenSecretRef: example-token
+  source:
+    apiGroup: "snapshot.kubevirt.io"
+    kind: VirtualMachineSnapshot
+    name: example-vmsnapshot
+```
+
+When you create a VMExport based on a Virtual Machine Snapshot, the controller will attempt to create PVCs from the volume snapshots contained in Virtual Machine Snapshot. Once all the PVCs are ready, the export server will start and you can begin the export. If the Virtual Machine Snapshot contains multiple volumes that can be exported, each volume will get its own URL links. If the Virtual Machine snapshot contains no volumes that can be exported, the VMExport will go into a `skipped` phase, and no export server is started.
+
+### Export Persistent Volume Claim
+
+After you have created the token you can now create a VMExport CR that identifies the Persistent Volume Claim (PVC) you want to export. You can create a VMExport that looks like this:
 
 ```yaml
 apiVersion: export.kubevirt.io/v1alpha1
@@ -27,10 +80,18 @@ spec:
   source:
     apiGroup: ""
     kind: PersistentVolumeClaim
-    name: vm-disk
+    name: example-pvc
 ```
 
-In this example the PVC name is vm-disk, but it will be different for your Virtual Machine Disk. Note the PVC doesn't need to contain a Virtual Machine Disk, it can contain any content, but the main use case is exporting Virtual Machine Disks. After you post this yaml to the cluster, and the PVC is not in use by another pod, a new export server is created in the same namespace as the PVC. The VirtualMachineExport CR will contain a status with internal and external links to the export service. The internal links are only valid inside the cluster, and the external links are valid for external access. The `cert` field will contain the CA that signed the certificate of the export server for internal links, or the CA that signed the Route or Ingress.
+In this example the PVC name is `example-pvc`. Note the PVC doesn't need to contain a Virtual Machine Disk, it can contain any content, but the main use case is exporting Virtual Machine Disks. After you post this yaml to the cluster, a new export server is created in the same namespace as the PVC. If the source PVC is in use by another pod (such as the virt-launcher pod) then the export will remain pending until the PVC is no longer in use. If the exporter server is active and another pod starts using the PVC, the exporter server will be terminated until the PVC is not in use anymore.
+
+
+### Export status links
+
+The VirtualMachineExport CR will contain a status with internal and external links to the export service. The internal links are only valid inside the cluster, and the external links are valid for external access through an Ingress or Route. The `cert` field will contain the CA that signed the certificate of the export server for internal links, or the CA that signed the Route or Ingress.
+
+#### KubeVirt content-type
+Example of exporting a PVC that contains a KubeVirt disk image. The controller determines if the PVC contains a kubevirt disk by checking if there is a special annotation on the PVC, or if there is a DataVolume ownerReference on the PVC, or if the PVC has a volumeMode of block.
 
 ```yaml
 apiVersion: export.kubevirt.io/v1alpha1
@@ -42,7 +103,7 @@ spec:
   source:
     apiGroup: ""
     kind: PersistentVolumeClaim
-    name: example-disk
+    name: example-pvc
   tokenSecretRef: example-token
 status:
   conditions:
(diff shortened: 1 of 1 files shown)


jpeimer (commented):
Thanks a lot for the doc! 
Only a couple of tiny suggestions.

viveksinghggits:
Hi @awels 
Shouldn't we have steps to import it in the cluster, as well. Or thats planned as part of another PR.

awels:
So the export should be compatible with CDI, so once you have the external URLs, you should be able to point CDI to it, and that will copy it into the cluster CDI is in.

viveksinghggits:
I was looking at the description of this PR https://github.com/kubevirt/kubevirt/pull/7863 and it looks  like to be able to import the exported PVC in another cluster we must have connectivity with the first cluster. Is that true? If yes, what if we don't have connectivity between them?

awels:
If you don't have connectivity between the clusters, how are you expecting the data to go from one cluster to another? Some kind of intermediate storage?

viveksinghggits:
mm hmm, but we can  not configure `VirtualMachineExport` to upload the disk image somewhere (the place CDI will be pointing to) else right? It will just be present in the source cluster (I am assuming).
Once we can get the disk image exported on a central location then we can utilise CDI I think easily.

viveksinghggits:
The other question that I had is, how do you recommend getting the VirtualMachine manifest on the other cluster, should we just export the VirtualMachine manifest from source cluster or `VirtualMachineSnapshotContent`.

awels:
> mm hmm, but we can not configure `VirtualMachineExport` to upload the disk image somewhere (the place CDI will be pointing to) else right? It will just be present in the source cluster (I am assuming). Once we can get the disk image exported on a central location then we can utilise CDI I think easily.

Exactly CDI will be used in the target cluster.

awels:
The manifest is the easy part, you can just export it with kubectl or whatever tool you use, and then import it into the target cluster. Might have to modify the names of the volumes if they changed in the new cluster. We are discussing some format for exporting an entire VMs including everything into a single format. But haven't reached a conclusion yet.

viveksinghggits:
> > mm hmm, but we can not configure `VirtualMachineExport` to upload the disk image somewhere (the place CDI will be pointing to) else right? It will just be present in the source cluster (I am assuming). Once we can get the disk image exported on a central location then we can utilise CDI I think easily.
> 
> Exactly CDI will be used in the target cluster.


Sorry, I still have some questions about this. 
Creation of the `VirtualMachineExport` resource would give me a publicly accessible link for the disk image (the snapshot would still be present in the source cluster). But if we don't have connectivity between source and target cluster, I think, right now we don't have a way to get the snapshot exported to the target cluster.
Because even if can use CDI to download the data from s3, how are we going to get the data (snapshot) to s3.

awels:
You would need something that has access to the source cluster URL and the S3 bucket, and it downloads it from the source cluster, and writes it to the s3 bucket. Then CDI can pull it from the S3 bucket.

viveksinghggits:
> You would need something that has access to the source cluster URL and the S3 bucket, and it downloads it from the source cluster, and writes it to the s3 bucket. Then CDI can pull it from the S3 bucket.

Makes sense, thank you. :pray:

docs/operations/export_api.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Export Virtual Machine disk
+
+After you have created the token you can now create a VMExport CR that identifies the disk of the VM you want to export. In KubeVirt in the end all persisted Virtual Machine disks are stored in a Persistent Volume Claim (PVC) in the namespace the Virtual Machine resides in. Once you have identified the PVC you can create a VMExport that looks like this:
+
+```yaml
+apiVersion: export.kubevirt.io/v1alpha1
+kind: VirtualMachineExport
+metadata:
+    name: example-export
```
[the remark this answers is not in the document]
awels: Done

docs/operations/export_api.md
```diff
@@ -0,0 +1,97 @@
+# Export API
+
+## Prerequesites
+None
```
ShellyKa13, line 4: why None if the feature gate must be enabled? isnt it a Prerequesite?

docs/operations/export_api.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Export Feature Gate
+
+VMExport support must be enabled in the feature gates to be supported. The
+[feature gates](./activating_feature_gates.md#how-to-activate-a-feature-gate)
+field in the KubeVirt CR must be expanded by adding the `VMExport` to it.
+
+### Export token
+
+In order to securely export a Virtual Machine Disk, you must create a token that is used to authorize users accessing the export endpoint. This token must be in a secret that is stored in the namespace the Virtual Machine resides in. The contents of the secret can be passed as a token header or parameter to the export URL. The name of the header or argument is `x-kubevirt-export-token` with a value that matches the content of the secret. The secret can be named any valid secret in the namespace. We recommend you use best security practices when generating this token.
```
ShellyKa13, line 14: maybe add a yaml for example?

docs/operations/export_api.md
```diff
@@ -0,0 +1,97 @@
...
+  name: example-export
+spec:
+  tokenSecretRef: example-token
+  source:
+    apiGroup: ""
+    kind: PersistentVolumeClaim
+    name: vm-disk
+```
+
+In this example the PVC name is vm-disk, but it will be different for your Virtual Machine Disk. Note the PVC doesn't need to contain a Virtual Machine Disk, it can contain any content, but the main use case is exporting Virtual Machine Disks. After you post this yaml to the cluster, and the PVC is not in use by another pod, a new export server is created in the same namespace as the PVC. The VirtualMachineExport CR will contain a status with internal and external links to the export service. The internal links are only valid inside the cluster, and the external links are valid for external access. The `cert` field will contain the CA that signed the certificate of the export server for internal links, or the CA that signed the Route or Ingress.
```
ShellyKa13, line 33: maybe hightlight **not in use by another pod**

docs/operations/export_api.md
```diff
@@ -0,0 +1,97 @@
...
+          url: https://virt-export-example-export.example.svc/volumes/example-disk/disk.img
+        - format: gzip
+          url: https://virt-export-example-export.example.svc/volumes/example-disk/disk.img.gz
+        name: example-disk
+  phase: Ready
+  serviceName: virt-export-example-export
+```
+
+If the source PVC is in use by another pod (such as the virt-launcher pod) then the export will remain pending until the PVC is no longer in use.
+
```
ShellyKa13, line 92: An example for how to use the created links is welcomed, maybe aexample to use it as source for new dv? and download to laptop?

ShellyKa13 (commented):
looks good, but I think more examples and how to use will add a lot

docs/operations/export_api.md
awels: Because featuregates are a separate item in the documentation. So besides the featuregate, there are none.

docs/operations/export_api.md
awels: Its just creating a secret, the 'hard' part is the generation of the token itself. I can add a yaml to create a secret but it is not very interesting.

docs/operations/export_api.md
```diff
@@ -0,0 +1,425 @@
+# Export API
+
+## Prerequesites
+None
+
+### Export Feature Gate
+
+VMExport support must be enabled in the feature gates to be supported. The
```
aglitke, line 8: How about:

VMExport must be enabled in the feature gates for this feature to be available

There is no "support" for any features in an open source project.

docs/operations/export_api.md
```diff
@@ -0,0 +1,425 @@
...
+[feature gates](./activating_feature_gates.md#how-to-activate-a-feature-gate)
+field in the KubeVirt CR must be expanded by adding the `VMExport` to it.
+
+### Export token
+
+In order to securely export a Virtual Machine Disk, you must create a token that is used to authorize users accessing the export endpoint. This token must be in a secret that is stored in the namespace the Virtual Machine resides in. The contents of the secret can be passed as a token header or parameter to the export URL. The name of the header or argument is `x-kubevirt-export-token` with a value that matches the content of the secret. The secret can be named any valid secret in the namespace. We recommend you generate an alpha numeric token of at least 12 characters. The data key should be `token`. For example:
```
aglitke, line 14: in the namespace where the Virtual Machine resides.

docs/operations/export_api.md
```diff
@@ -0,0 +1,425 @@
...
+    name: example-vm
+```
+
+The following volumes present in the VM will be exported:
+
+* PersistentVolumeClaims
+* DataVolumes
+* MemoryDump
+
+All other volume types are not exported. The export will only be available when the VM is not running, this is to prevent corrupt data from being exported. If you want to export data from a running VM, the VM will have to use storage that can be snapshotted, and you can create a Virtual Machine Snapshot and export that while the VM stays running. If an export is active and the VM is started the export will detect this state and terminate the export server.
```
aglitke, line 46: To avoid the export of inconsistent data, a Virtual Machine can only be exported while it is powered off.  Any active VM exports will be terminated if the Virtual Machine is started.  To export data from a running Virtual Machine you must first create a Virtual Machine Snapshot (see below).

docs/operations/export_api.md
mhenriks: Yeah I think I'm with @ShellyKa13.  Seems like exactly the sort of thing that should go here.

awels: Well all the other pages have separate sections for enabling the feature gate. I think I am going to just remove the prerequesites section since it makes no sense.

docs/operations/export_api.md
awels: Done

docs/operations/export_api.md
awels: Done

docs/operations/export_api.md
awels: Done

docs/operations/export_api.md
awels: I added an example of how to copy an image from one cluster to another using CDI.

brybacki:
Nice, I really like the Use case. 

I'd also like to see an introductory first paragraph that simply describes what exactly the VMExport is, and why one might use it. Similar to the first paragraph in the hotplug document. https://github.com/kubevirt/user-guide/blob/main/docs/operations/hotplug_volumes.md

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
+# Export API
+
```
cwilkers, line 2: It would be nice here to have a bit of intro text so the reader can quickly scan from search results to see if this is what they're after.

Also, small nit, you go from H1 straight down to H3, bypassing H2... You could promote most of your headers by one or add some big idea chapter headings like Prerequisites.

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
...
+### Export Feature Gate
+
+VMExport support must be enabled in the feature gates to be available. The
+[feature gates](./activating_feature_gates.md#how-to-activate-a-feature-gate)
+field in the KubeVirt CR must be expanded by adding the `VMExport` to it.
+
+### Export token
+
+In order to securely export a Virtual Machine Disk, you must create a token that is used to authorize users accessing the export endpoint. This token must be in the namespace the Virtual Machine resides. The contents of the secret can be passed as a token header or parameter to the export URL. The name of the header or argument is `x-kubevirt-export-token` with a value that matches the content of the secret. The secret can be named any valid secret in the namespace. We recommend you generate an alpha numeric token of at least 12 characters. The data key should be `token`. For example:
```
cwilkers, line 11:
```suggestion
In order to securely export a Virtual Machine Disk, you must create a token that is used to authorize users accessing the export endpoint. This token must be in the same namespace as the Virtual Machine. The contents of the secret can be passed as a token header or parameter to the export URL. The name of the header or argument is `x-kubevirt-export-token` with a value that matches the content of the secret. The secret can be named any valid secret in the namespace. We recommend you generate an alpha numeric token of at least 12 characters. The data key should be `token`. For example:
```

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
...
+* DataVolumes
+* MemoryDump
+
+All other volume types are not exported. To avoid the export of inconsistent data, a Virtual Machine can only be exported while it is powered off. Any active VM exports will be terminated if the Virtual Machine is started. To export data from a running Virtual Machine you must first create a Virtual Machine Snapshot (see below).
+
+If the VM contains multiple volumes that can be exported, each volume will get its own URL links. If the VM contains no volumes that can be exported, the VMExport will go into a `Skipped` phase, and no export server is started.
+
+### Export Virtual Machine Snapshot volumes
+
+After you have created the token you can now create a VMExport CR that identifies the Virtual Machine Snapshot you want to export. You can create a VMExport that looks like this:
```
cwilkers, line 49: Writing "After you have created the token" in each subsection seems redundant. Perhaps you could add a paragraph above the various export types saying after you have created the token, you can now create one of the following N CRs depending on the data you would like to export.

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
...
+
+
+### Export status links
+
+The VirtualMachineExport CR will contain a status with internal and external links to the export service. The internal links are only valid inside the cluster, and the external links are valid for external access through an Ingress or Route. The `cert` field will contain the CA that signed the certificate of the export server for internal links, or the CA that signed the Route or Ingress.
+
+#### KubeVirt content-type
+Example of exporting a PVC that contains a KubeVirt disk image. The controller determines if the PVC contains a kubevirt disk by checking if there is a special annotation on the PVC, or if there is a DataVolume ownerReference on the PVC, or if the PVC has a volumeMode of block.
```
cwilkers, line 91: This starts with a sentence fragment; suggest "The following is an example..."

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
...
+        - format: raw
+          url: https://virt-export-example-export.example.svc/volumes/example-disk/disk.img
+        - format: gzip
+          url: https://virt-export-example-export.example.svc/volumes/example-disk/disk.img.gz
+        name: example-disk
+  phase: Ready
+  serviceName: virt-export-example-export
+```
+#### Archive content-type
+
```
cwilkers, line 146: Could use some explanation here, does "archive content-type" apply any time the content is determined not to be a kubevirt disk image?

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
...
+Raw and Gzip will be selected if the PVC is determined to be a KubeVirt disk. KubeVirt disks contain a single disk.img file (or are a block device). Dir will return a list of the files in the PVC, to download a specific file you can replace `/dir` in the URL with the path and file name. For instance if the PVC contains the file `/example/data.txt` you can replace `/dir` with `/example/data.txt` to download just data.txt file. Or you can use the tar.gz URL to get all the contents of the PVC in a tar file.
+
+
+#### Internal link certificates
+The export server certificate is valid for 7 days after which it is rotated by deleting the export server pod and associated secret and generating a new one. If for whatever reason the export server pod dies, the associated secret is also automatically deleted and a new pod and secret are generated. The VirtualMachineExport object status will be automatically updated to reflect the new certificate.
+
+#### External link certificates
+The external link certificates as associated with the Ingress/Route that points to the service created by the KubeVirt operator. The CA that signed the Ingress/Route will part of the certificates.
```
cwilkers, line 213:
```suggestion
The external link certificates are associated with the Ingress/Route that points to the service created by the KubeVirt operator. The CA that signed the Ingress/Route will be part of the certificates. 
```

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
...
+metadata:
+  name: router-cert
+data:
+  ca.pem: |
+    -----BEGIN CERTIFICATE-----
+    ...
+    -----END CERTIFICATE-----
+```
+
+Next create a secret in the same namespace on the target cluster, note the token value will be a header that is passed to the server by CDI.
```
cwilkers, line 361:
```suggestion
Next, create a secret in the same namespace on the target cluster. Note: the token value will be a header that is passed to the server by CDI.
```

docs/operations/export_api.md
```diff
@@ -0,0 +1,422 @@
...
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: secret-headers
+stringData:
+  token: "x-kubevirt-export-token:1234567890ab"
+```
+Note, make sure there is no ` ` between the `:` and the actual token, otherwise the space is sent as part of the header, the authentication will fail.
+Now we can go ahead and import the disk image into the target cluster using a data volume, for convenience I put the data volume inside a data volume template section of the same VM spec as the source:
```
cwilkers, line 372:
```suggestion
Now we can go ahead and import the disk image into the target cluster using a data volume. For convenience, I put the data volume inside a data volume template section of the same VM spec as the source:
```

cwilkers (changes requested):
Thanks for the PR, this is a great feature!

I have included some readability suggestions

docs/operations/export_api.md
awels: Reason I go from H1 to H3 is that H2 renders almost the same as H1 and it is hard to tell the difference between them. I can definitely add some intro text to explain the point of this.

awels:
Should have addressed all the review comments.

docs/operations/export_api.md
```diff
@@ -0,0 +1,426 @@
...
+          url: https://virt-export-example-export.example.svc/volumes/example/dir
+        - format: tar.gz
+          url: https://virt-export-example-export.example.svc/volumes/example/disk.tar.gz
+        name: example-disk
+  phase: Ready
+  serviceName: virt-export-example-export
+```
+#### Format types
+There are 4 format types that are possible:
+* Raw. The unaltered raw KubeVirt disk image.
```
cwilkers, line 205: Looking at the rendered copy, the following bullets did not render properly because there has to be a newline before a list.

```suggestion

* Raw. The unaltered raw KubeVirt disk image.
```

cwilkers (changes requested):
One formatting error, outside of that all is well.

docs/operations/export_api.md
awels: Weird my local renderer renders it properly. Added the newline.

awels:
@cwilkers you good with this now?

cwilkers (approved):
All good!

/lgtm
/approve

PR #554, opened 2022-06-30
Certificate for virtctl image-upload can be used without update-ca-trust

jpeimer:
For Fedora/RHEL:
When you add a certificate to "/etc/pki/tls/certs", you don't need to run 'sudo update-ca-trust'.

### DIFF
--- a/docs/operations/containerized_data_importer.md
+++ b/docs/operations/containerized_data_importer.md
@@ -198,9 +198,7 @@ Download the cdi-uploadproxy-server-cert.
 
 Add this certificate to the systems trust store. On Fedora, this can be done as follows.
 
-    sudo cp cdi-uploadproxy-server-cert.crt /etc/pki/ca-trust/source/anchors
-
-    sudo update-ca-trust
+    sudo cp cdi-uploadproxy-server-cert.crt /etc/pki/tls/certs
 
 The upload should now work.
 


jpeimer:
Signed-off-by: Jenia Peimer [jpeimer@redhat.com](mailto:jpeimer@redhat.com)

cwilkers:
It may work to put certs in /etc/pki/tls/certs for some applications (including the web browser involved in this documentation) but it isn't the official system trust store per https://docs.fedoraproject.org/en-US/quick-docs/using-shared-system-certificates/.
I would hesitate to change the documentation because the functionality you're relying on is a side effect of the browser supporting the old, deprecated way of adding certificates, and it is conceivable this could go away with future releases (or with different browsers).

Caveat: I could not easily find documentation to the effect that /etc/pki/tls is legacy / deprecated, but the update-ca-trust man page mentions that applications expecting to find certificates in a fixed location are "legacy"

jpeimer:
Hi Chandler, thanks for the review. That's where we found out about 'etc/pki/tls': ​https://go.dev/src/crypto/x509/root_linux.go

So it can be used if we don't have sudo permissions for update-ca-trust. But update-ca-trust looks like a better way anyway, so I'm closing this PR for now.

PR #559, opened 2022-07-11
Adjust on the get of the latest CDI tag version

rafaeldomi:
The current curl don't get the latest CDI version. The proposal is to read the "location" that curl receives in the 302 return of the request, and parses to get the tag in the URL.

Signed-off-by: Rafael Domiciano <rafael.domiciano@gmail.com>

### DIFF
--- a/docs/operations/containerized_data_importer.md
+++ b/docs/operations/containerized_data_importer.md
@@ -21,7 +21,8 @@ virtctl in your path.
 Install the latest CDI release
 [here](https://github.com/kubevirt/containerized-data-importer/releases)
 
-    VERSION=$(curl -s https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
+    export TAG=$(curl -s -w %{redirect_url} https://github.com/kubevirt/containerized-data-importer/releases/latest)
+    export VERSION=$(echo ${TAG##*/})
     kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-operator.yaml
     kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-cr.yaml
 


docs/operations/containerized_data_importer.md
```diff
@@ -21,7 +21,8 @@ virtctl in your path.
 Install the latest CDI release
 [here](https://github.com/kubevirt/containerized-data-importer/releases)
 
-    VERSION=$(curl -s https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
+    export TAG=$(curl -s -w %{redirect_url} https://github.com/kubevirt/containerized-data-importer/releases/latest)
```
awels, line 24: Note this is the bash function we use in the code to get the latest you just call `get_latest_release "kubevirt/kubevirt" and it finds the latest release.

```
get_latest_release() {
  curl -s "https://api.github.com/repos/$1/releases/latest" |       # Get latest release from GitHub api
    grep '"tag_name":' |                                            # Get tag line
    sed -E 's/.*"([^"]+)".*/\1/'                                    # Pluck JSON value (avoid jq)
}
```
Now I don't mind your version, but it might make sense to show the same as the actual project.

cwilkers: @awels we use the above curl | grep -o structure in multiple places in the docs (and labs, and scenarios), I really like the %{redirect_url} trick, and would consider that more future safe than a grep of a generated page...

Once we decide which of the two methodologies is best, @rafaeldomi would you be open to creating a similar PR against https://github.com/kubevirt/kubevirt.github.io ?

cwilkers: Also, a one-liner way (if needed) would be to nest $() executions like:

```
export VERSION=$(basename $(curl -s -w %{redirect_url} https://github.com/kubevirt/containerized-data-importer/releases/latest))
```

awels: Note I am perfectly fine with what is shown here.

cwilkers: > Note I am perfectly fine with what is shown here.

As am I

/approve

PR #566, opened 2022-07-25
Update memory dump doc with new feature improvements

ShellyKa13:
This PR links to another documentation which is on a PR so need it to be merged first: https://github.com/kubevirt/user-guide/pull/549
Signed-off-by: Shelly Kagan <skagan@redhat.com>

### DIFF
--- a/docs/operations/memory_dump.md
+++ b/docs/operations/memory_dump.md
@@ -21,7 +21,8 @@ Now lets assume we have a running VM and the name of the VM is 'my-vm'.
 We can either dump to an existing pvc, or request one to be created.
 
 #### Existing PVC
-The size of the PVC must be big enough to hold the memory dump (The VM memory size + 100Mi overhead multiplied with the filesystem overhead),
+The size of the PVC must be big enough to hold the memory dump. The calculation is (VMMemorySize + 100Mi) * FileSystemOverhead, Where `VMMemorySize` is the memory size,
+100Mi is reserved space for the memory dump overhead and `FileSystemOverhead` is the value used to adjust requested PVC size with the filesystem overhead.
 also the PVC must have a `FileSystem` volume mode.
 
 Example for such PVC:
--- a/docs/operations/memory_dump.md
+++ b/docs/operations/memory_dump.md
@@ -13,9 +13,16 @@ The memory dump process mounts a PVC to the virt-launcher in order to get the ou
 [feature gates](./activating_feature_gates.md#how-to-activate-a-feature-gate)
 field in the KubeVirt CR must be expanded by adding the `HotplugVolumes` to it.
 
-### A PVC to hold the memory dump output
+## Virtctl support
 
-Currently in order to get the memory dump output a pre-existing PVC is required. The size of the PVC must be big enough to hold the memory dump (The VM memory size + 100Mi overhead).
+### Get memory dump
+
+Now lets assume we have a running VM and the name of the VM is 'my-vm'.
+We can either dump to an existing pvc, or request one to be created.
+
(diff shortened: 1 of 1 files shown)


docs/operations/memory_dump.md
```diff
@@ -13,9 +13,16 @@ The memory dump process mounts a PVC to the virt-launcher in order to get the ou
...
+## Virtctl support
+
+### Get memory dump
+
+Now lets assume we have a running VM and the name of the VM is 'my-vm'.
+We can either dump to an existing pvc, or request one to be created.
 
-Currently in order to get the memory dump output a pre-existing PVC is required. The size of the PVC must be big enough to hold the memory dump (The VM memory size + 100Mi overhead).
+#### Existing PVC
+The size of the PVC must be big enough to hold the memory dump (The VM memory size + 100Mi overhead multiplied with the filesystem overhead),
```
brybacki, line 24: this `The VM memory size + 100Mi overhead multiplied with the filesystem overhead,` might be a little bit ambiguous. I think I know what it means, but I can also read it as `memSize + (100Mi * fsOverhead)` or (memSize +  100Mi) * fsOverhead. 

And which one is correct? And why 100Mi?

docs/operations/memory_dump.md
```diff
@@ -13,9 +13,16 @@ The memory dump process mounts a PVC to the virt-launcher in order to get the ou
 [feature gates](./activating_feature_gates.md#how-to-activate-a-feature-gate)
 field in the KubeVirt CR must be expanded by adding the `HotplugVolumes` to it.
 
-### A PVC to hold the memory dump output
+## Virtctl support
+
+### Get memory dump
+
+Now lets assume we have a running VM and the name of the VM is 'my-vm'.
+We can either dump to an existing pvc, or request one to be created.
```
brybacki, line 21: I think it looks and reads better with "You" instead of "We" in documentation.

But I am not a native speaker, so do take my advice with a grain of salt.

docs/operations/memory_dump.md
ShellyKa13: the correct one is 

> (memSize + 100Mi) * fsOverhead

100Mi is a const overhead of the memorydump there is no actual explanation.. added of some headers and stuff..
How would you think is best writing so it will be clear.. its already in braces () so should I write what you did? maybe without the fs and mem? i.e:
( (VMMemorySize+100Mi) * FilesystemOverhead )
I wanted to put it more in words then in a formula but if its not clear then ok.

docs/operations/memory_dump.md
ShellyKa13: yeah I thought about it, the 'lets assume we have' I took from another documentation and (also in this PR I just moved it, it is already in the original doc). Since the 'lets assume it made me write the entire doc in 'we' instead of 'you'.
I'll try to rephrase..

docs/operations/memory_dump.md
brybacki: Maybe something like this, if not too verbose. Just change *stuff* with some other word that better describes it, are these logs or utils? 

The size of the PVC must be big enough to hold the memory dump. The calculation is `( (VMMemorySize+ MemoryDumpOverhead) * FilesystemOverhead )`, where `VMMemorySize` is the size of VM memory, `MemoryDumpMetadata` is the space reserved for memory dump headers and some *stuff*, `FilesystemOverhead` is the value used to adjust requested PVC size withe filesystem overhead.

ShellyKa13: @brybacki added a new commit to rephrase this explanation

PR #570, opened 2022-08-02
passt documentation

AlonaKaplan:
This PR is adding passt documentation explaining how to run a VM with passt interface.
It also updated the istio documentation to mention istio is now supported not only by masquerade VMs but by passt as well.

### DIFF
--- a/docs/virtual_machines/interfaces_and_networks.md
+++ b/docs/virtual_machines/interfaces_and_networks.md
@@ -595,6 +595,67 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
 reaching a destination using its FQDN is not possible.
 Tracking issue - https://github.com/kubevirt/kubevirt/issues/7184
 
+### passt
+
+`passt` is a new approach for user-mode networking which can be used as a simple replacement for Slirp (which is practically dead).
+
+`passt` is a universal tool which implements a translation layer between a Layer-2 network interface and native
+Layer -4 sockets (TCP, UDP, ICMP/ICMPv6 echo) on a host.<br/>
+
+Its main benefits are:
+- doesn't require extra network capabilities as CAP_NET_RAW and CAP_NET_ADMIN.
+- allows integration with service meshes (which expect applications to run locally) out of the box.
+- supports IPv6 out of the box (in contrast to the existing bindings which require configuring IPv6
+manually).
+
+|                                                                | Masquerade                                                                                             | Bridge                                                                    | Passt                                                                                                   |
+|----------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------|
+| Supports migration                                             | Yes                                                                                                    | No                                                                        | No<br/>(will be supported in the future)                                                                |
+| VM uses Pod IP                                                 | No                                                                                                     | Yes                                                                       | Yes<br/>(in the future it will be possible to configure the VM IP. Currently the default is the pod IP) |
+| Service Mesh out of the box                                    | No<br/>(only ISTIO is supported, adjustmets on both ISTIO and kubevirt had to be done to make it work) | No                                                                        | Yes                                                                                                     |
+| Doesn’t require extra capabilities on the virt-launcher pod    | Yes<br/>(multiple workarounds had to be added to kuebivrt to make it work)                             | No<br/>(Multiple workarounds had to be added to kuebivrt to make it work) | Yes                                                                                                     |
+| Doesn't require extra network devices on the virt-launcher pod | No<br/>(bridge and tap device are created)                                                             | No<br/>(bridge and tap device are created)                                | Yes                                                                                                     |
+| Supports IPv6                                                  | Yes<br/>(requires manual configuration on the VM)                                                      | No                                                                        | Yes                                                                                                     |
+
+```yaml
+kind: VM
+spec:
+  domain:
+    devices:
+      interfaces:
+        - name: red
+          passt: {} # connect using passt mode
+          ports:
+            - port: 8080 # allow incoming traffic on port 8080 to get into the virtual machine
+  networks:
+    - name: red
+      pod: {}
+```
+
+#### Requirements/Recommendations:
+1. To get better performance the node should be configured with:
+```
+sysctl -w net.core.rmem_max = 33554432
+sysctl -w net.core.wmem_max = 33554432
+```
+2. To run multiple passt VMs with no explicit ports, the node's `fs.file_max` should be increased
+   (for a VM forwards all IPv4 and IPv6 ports, for TCP and UDP, passt needs to create ~2^18 sockets):
+```
+sysctl -w fs.file_max = 9223372036854775807
+```
+3. The `vmi.Spec.Domain.Resources.Requests.Memory` of a passt VM with no explicit ports should be at least 2048M.
+
+#### Temporary restrictions: 
+1. `passt` currently only supported as primary network and doesn't allow extra multus networks to be configured on the VM.
+2. A VM with passt binding cannot bind to low ports (less than 1024).
+
+passt interfaces are feature gated; to enable the feature, follow
+[these](../operations/activating_feature_gates.md#how-to-activate-a-feature-gate)
+instructions, in order to activate the `Passt` feature gate (case sensitive).
+
+More information about passt mode can be found in [passt
+Wiki](https://passt.top/passt/about/).
+
 ### virtio-net multiqueue
 
 Setting the `networkInterfaceMultiqueue` to `true` will enable the
--- a/docs/virtual_machines/istio_service_mesh.md
+++ b/docs/virtual_machines/istio_service_mesh.md
@@ -5,7 +5,7 @@ Kubevirt supports running VMs as a part of Istio service mesh.
 
 ## Limitations
 
-- Istio service mesh is only supported with a pod network masquerade binding.
+- Istio service mesh is only supported with a pod network masquerade or passt binding.
 
 - Istio uses a [list of ports](https://istio.io/latest/docs/ops/deployment/requirements/#ports-used-by-istio) for its own purposes, these ports must not be explicitly specified in a VMI interface.
 
@@ -126,7 +126,7 @@ $ kubectl describe pod virt-launcher-vmi-istio-lg5gp
   Warning  Unhealthy  2d8h (x4 over 2d8h)  kubelet            Readiness probe failed: Get "http://10.244.186.222:15021/healthz/ready": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
 ```
 
-**Resolution:** Make sure the `sidecar.istio.io/inject: "true"` annotation is defined in the created VMI and that masquerade binding is used for pod network interface.
+**Resolution:** Make sure the `sidecar.istio.io/inject: "true"` annotation is defined in the created VMI and that masquerade or passt binding is used for pod network interface.
 
 ### Virt-launcher pod for VMI is stuck at initialization phase
 ```shell


docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+    pod: {}
+```
+
+#### Requirements:
+1. To get better performance the node should be configured with:
+```
+sysctl -w net.core.rmem_max = 33554432
+sysctl -w net.core.wmem_max = 33554432
+```
+2. To run a multiple passt VMs with no explicit ports, the node's `fs.file_max` should be increased
```
rhrazdil, line 621:
```suggestion
2. To run multiple passt VMs with no explicit ports, the node's `fs.file_max` should be increased
```

What if user fails to do this? Is it worth adding a "debugging" section that would help users to easily identify that this must be increased?
Or maybe just a note saying tha if `this` happens, it's because of `that`

AlonaKaplan: If the user doesn't specify it, he will fail running the VM.

AlonaKaplan: I don't want to specify here the exact error the user will get since it is planed to add a validation webhook making sure the are enough free FDs with an explicit error.

rhrazdil: I see, that sound better than my suggestion :)

rhrazdil (approved):

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
 reaching a destination using its FQDN is not possible.
 Tracking issue - https://github.com/kubevirt/kubevirt/issues/7184
 
+### passt
+
+`passt` is a new approach for user-mode network that can be used as a simple replacement for Slirp.
```
phoracek, line 600: Can it be used as a valid replacement for the bridge binding on pod network too?

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+
+```yaml
+kind: VM
+spec:
+  domain:
+    devices:
+      interfaces:
```
phoracek, line 607: Could you show how to request explicit ports?

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+   (at least ~2^18 per passt with no explicit ports VM):
+```
+sysctl -w fs.file_max = 9223372036854775807
+```
+3. The `vmi.Spec.Domain.Resources.Requests.Memory` of a passt VM with no explicit ports should be at least 2048M.
+
+> **Temporary restrictions:** 
+> 1. `passt` currently only supported as primary network
+> and doesn't allow extra multus networks to be configured on the VM.
+> 2. A VM with passt binding cannot bind to low ports.
```
phoracek, line 631: Could you define the range?

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+    pod: {}
+```
+
+#### Requirements:
+1. To get better performance the node should be configured with:
+```
+sysctl -w net.core.rmem_max = 33554432
+sysctl -w net.core.wmem_max = 33554432
+```
+2. To run multiple passt VMs with no explicit ports, the node's `fs.file_max` should be increased
```
phoracek, line 621: Could you note how much RAM one VM with all ports open takes? And how much for a VM requesting a single port. To motivate users to be specific

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+   (at least ~2^18 per passt with no explicit ports VM):
+```
+sysctl -w fs.file_max = 9223372036854775807
```
phoracek, line 624: How many VMs this allows for?

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+```
+3. The `vmi.Spec.Domain.Resources.Requests.Memory` of a passt VM with no explicit ports should be at least 2048M.
```
phoracek, line 626: Is this at all related to spec.domain.memory.guest? If I set both to the same value, all the memory would be given to the guest, and you would not have enough left for passt running in launcher?

phoracek (changes requested):

docs/virtual_machines/interfaces_and_networks.md
maiqueb: hm, afaiu, virt-controller calculates the launcher overhead (some 100MB, depending on the "requested" guest features) and adds those to what you request here - the intention being assuring the guest can operate properly (while setting this threshold at the pod level).

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,48 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+
+> **Temporary restrictions:** 
+> 1. `passt` currently only supported as primary network
+> and doesn't allow extra multus networks to be configured on the VM.
```
maiqueb, line 630: Could you elaborate on the reasoning behind this ? 

Not sure that makes sense on the document: I'm just curious - is it because passt is **very** hungry for memory ?

maiqueb (changes requested):
I think more info is required.

docs/virtual_machines/interfaces_and_networks.md
AlonaKaplan: I believe Miguel is right and the memory requested from kubevirt when running a VM is global (not pod part and guest part).
`spec.domain.memory.guest` is used for [over-commitment](https://kubevirt.io/user-guide/operations/node_overcommit/)

docs/virtual_machines/interfaces_and_networks.md
AlonaKaplan: Regarding the exact RAM gap. I don't know,
But it is mentioned in the doc that ``vmi.Spec.Domain.Resources.Requests.Memory` should be set to at least `2048` when running passt VM with no explicit ports.
For "regular" VMs we usually request 1024.

@phoracek it you think more precise values are needed, it would require further investigation.
Can be done as a follow-up.

maiqueb: > Regarding the exact RAM gap. I don't know, But it is mentioned in the doc that ``vmi.Spec.Domain.Resources.Requests.Memory`should be set to at least`2048` when running passt VM with no explicit ports. For "regular" VMs we usually request 1024.
> 
> @phoracek it you think more precise values are needed, it would require further investigation. Can be done as a follow-up.

I agree on the follow-up ... This is highly experimental, and users can always open up issues requesting us to clarify some parts.

docs/virtual_machines/interfaces_and_networks.md
maiqueb: @AlonaKaplan maybe its just me, Could you clarify what this 2^18 magic number actually is ?

docs/virtual_machines/interfaces_and_networks.md
maiqueb: but would that require more memory for each of the pods or not ?... the question may actually be: would we run 1 passt process per iface (as we kind of do for the dhcp server - 1 thread ...) or, would 1 process be able to service all of them ?

This question is purely for my information: the doc is good as is.

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,66 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+|----------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
+| Supports migration                                             | V                                                                                                     | X                                                                        | X<br/>(will be supported in the future)                                                              |
+| Inner and outer IPs are the same                               | X                                                                                                     | V                                                                        | V<br/>(in the future it will be possible to configure the VM IP. Currently the default is the pod IP) |
```
maiqueb, line 613: Not sure about inner / outer IPs. I think it would be better to be explicit, and refer what they actually are: pod / VM IPs.

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,66 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
 reaching a destination using its FQDN is not possible.
 Tracking issue - https://github.com/kubevirt/kubevirt/issues/7184
 
+### passt
+
+`passt` is a new approach for user-mode networking which can be used as a simple replacement for Slirp (which is practically dead).
```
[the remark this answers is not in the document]
maiqueb:
```suggestion
`passt` is a new approach for user-space networking which can be used as a simple replacement for Slirp (which is practically dead).
```
Not sure I'm being too picky here ... But afaiu, that's the common "term" .

maiqueb (changes requested):
Thanks for the table. I wanted that, but wasn't even aware.

Minor cosmetic changes requested.

docs/virtual_machines/interfaces_and_networks.md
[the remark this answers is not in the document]
AlonaKaplan: Copied it from the passt doc, prefer to keep it consistent.

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,66 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+
+`passt` is a universal tool which implements a translation layer between a Layer-2 network interface and native
+Layer -4 sockets (TCP, UDP, ICMP/ICMPv6 echo) on a host.<br/>
+The main benefits of `passt` are that it doesn't require extra network capabilities as CAP_NET_RAW and CAP_NET_ADMIN
+and allows integration with service meshes (which expect applications to run locally) out of the box.
+
+Moreover, `passt` supports IPv6 out of the box (in contrast to the existing bindings which require configuring IPv6
+manually).
```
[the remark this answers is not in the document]
AlonaKaplan: done

docs/virtual_machines/interfaces_and_networks.md
AlonaKaplan: done.

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -595,6 +595,66 @@ has no DHCP server at all. Therefore, the VM won't have the search domains infor
...
+| Doesn’t require extra capabilities on the virt-launcher pod    | V<br/>(multiple workarounds had to be added to kuebivrt to make it work)                              | X<br/>(Multiple workarounds had to be added to kuebivrt to make it work) | V                                                                                                    |
+| Doesn't require extra network devices on the virt-launcher pod | X<br/>(bridge and tap device are created)                                                             | X<br/>(bridge and tap device are created)                                | V                                                                                                    |
+| Supports IPv6                                                  | V<br/>(requires manual configuration on the VM)                                                       | X                                                                        | V                                                                                                    |
```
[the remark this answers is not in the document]
AlonaKaplan: done.

docs/virtual_machines/interfaces_and_networks.md
AlonaKaplan: 2^16 is the max port range. You need to open a socket per port (TCP IPv4, TCP IPv6, UDP IPv4, UDP IPv6). So 2^16 *4.
Added this ` for a VM forwards all IPv4 and IPv6 ports, for TCP and UDP, passt needs to create ~2^18 sockets` to the doc. I don't think the exact calculation is needed here.

maiqueb: Thanks for explaining.

maiqueb (approved):
Thank you.

phoracek:
/approve

Thanks for covering this

PR #572, opened 2022-08-12
Fix CDI version pulling

kissiel:
The cURL invocation did not follow the redirect yielding no version. Added -L to follow links. Also added '-m1' to grep so only the first match is used instead of a long string like "v1.53.0 v1.53.0 v1.53.0 v1.52.0 v1.53.0 v1.52.0 v1.53.0..."

### DIFF
--- a/docs/operations/containerized_data_importer.md
+++ b/docs/operations/containerized_data_importer.md
@@ -21,7 +21,7 @@ virtctl in your path.
 Install the latest CDI release
 [here](https://github.com/kubevirt/containerized-data-importer/releases)
 
-    VERSION=$(curl -s https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
+    VERSION=$(curl -sL https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -m1 -o "v[0-9]\.[0-9]*\.[0-9]*")
     kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-operator.yaml
     kubectl create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$VERSION/cdi-cr.yaml
 


cwilkers:
Thanks for your PR!

/ok-to-test

docs/operations/containerized_data_importer.md
```diff
@@ -21,7 +21,7 @@ virtctl in your path.
 Install the latest CDI release
 [here](https://github.com/kubevirt/containerized-data-importer/releases)
 
-    VERSION=$(curl -s https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
+    VERSION=$(curl -sL https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -m1 -o "v[0-9]\.[0-9]*\.[0-9]*")
```
jean-edouard, line 24: This seems to be the globally accepted way of finding the latest release of a github project...
However, it's effectively grepping the source of an entire webpage for a not-so-specific regex, which is not so reliable.
I mean, this is a random line from the source code, which comes *so close* to matching the regex:
```
<path fill-rule="evenodd" d="M0 6.75C0 5.784.784 5 1.75 5h1.5a.75.75 0 010 1.5h-1.5a.25.25 0 00-.25.25v7.5c0 .138.112.25.25.25h7.5a.25.25 0 00.25-.25v-1.5a.75.75 0 011.5 0v1.5A1.75 1.75 0 019.25 16h-7.5A1.75 1.75 0 010 14.25v-7.5z"></path><path fill-rule="evenodd" d="M5 1.75C5 .784 5.784 0 6.75 0h7.5C15.216 0 16 .784 16 1.75v7.5A1.75 1.75 0 0114.25 11h-7.5A1.75 1.75 0 015 9.25v-7.5zm1.75-.25a.25.25 0 00-.25.25v7.5c0 .138.112.25.25.25h7.5a.25.25 0 00.25-.25v-7.5a.25.25 0 00-.25-.25h-7.5z"></path>
```

Here's an alternative idea, let me know what you think!
```suggestion
    VERSION=$(curl -sL -o /dev/null -w %{url_effective} https://github.com/kubevirt/containerized-data-importer/releases/latest | xargs basename)
```

jean-edouard (commented):
Good catch, thank you for your contribution! See comment in-line.

cwilkers:
This seems to be a duplicate with https://github.com/kubevirt/user-guide/pull/559

fabiand:
@aglitke @awels what should we do about this one?

kissiel:
Hey folks, gimme 48h. I see if I can make it better than grep.
Sorry for the delay.

cwilkers:
No problem @kissiel but you should know that a similar PR got merged a while back, so this one may be unnecessary now (and definitely has merge conflicts with the current repo that need to be addressed first.)

kissiel:
> No problem @kissiel but you should know that a similar PR got merged a while back, so this one may be unnecessary now (and definitely has merge conflicts with the current repo that need to be addressed first.)

Ah, ok, so that solves it, lemme close this PR then.

PR #575, opened 2022-08-16
Update virtual_hardware.md

Barakmor1:
Update virtual_hardware.md
We stopped considering minCPUModel field in #7765
It might cause problems when migrating with
host-model CPU-Model.


jrcichra:
Could we also add a note about `node-labeller.kubevirt.io/skip-node=true`?

fossedihelm:
Thanks!
/lgtm

PR #578, opened 2022-08-30
Update the cloudInitNoCloud userdata example

farcaller:
userData should start with `#cloud-config` or the parsing will fail.


cwilkers:
Good catch, thanks for the PR.

Let's see how many instructions the bot will let me get away with all at once:

/ok-to-test
/approve
/lgtm

PR #580, opened 2022-09-04
fix typo in macvtap config

mikelsid:
Signed-off-by: Michael Sidorov <mixity@yandex.ru>
fixed typo in macvtap conig


cwilkers:
Thanks for the PR!

I'm not sure why the extra dashes in the table were important, but it doesn't seem to do any harm. Good catch on the missing comma.

/lgtm
/approve

PR #595, opened 2022-09-29
installation docs: add minimum container-selinux version

jean-edouard:


xpivarc (commented):
I think this will be hard to version so I would prefer to have this information in Kubevirt main repo. WDYT?

jean-edouard:
> I think this will be hard to version so I would prefer to have this information in Kubevirt main repo. WDYT?

That's a good point, do you have a specific location in mind?
What do you think about keeping this PR too with a link to that kubevirt/kubevirt file so people are made aware of it when they follow the docs?

jean-edouard:
Reworded this and added https://github.com/kubevirt/kubevirt/pull/8580.
Will close this PR if deemed redundant.
@xpivarc PTAL!

docs/operations/installation.md
```diff
@@ -41,6 +41,12 @@ virtualization workloads:
       QEMU: Checking if device /dev/net/tun exists                               : PASS
     ...
 
+### SELinux support
+
+SELinux-enabled nodes need [Container-selinux](https://github.com/containers/container-selinux) installed. The minimum version is documented inside the kubevirt/kubevirt repository, in docs/getting-started.md, under "SELinux support".
+
+For (older) release branches that don't specify a container-selinux version, version 2.170.0 or newer is required.
```
xpivarc, line 48: `required` -> `recommended`? I mean at least for me it is hard to say what we require in X(pick a random one :) ) without a deeper look.

xpivarc (commented):
/lgtm

cwilkers:
I think it's valuable to have the information in the user-guide pointing at the kubevirt/kubevirt docs. This will help searchability.

/approve

rmohr:
> I think it's valuable to have the information in the user-guide pointing at the kubevirt/kubevirt docs. This will help searchability.
> 
> /approve

+1 on that. If it is in the kubevirt main repo, then it is separate from the install guide which users will use. Also the release cadence is now much slower, decreasing the chance that this gets out of sync, since we have quite some time to update the user-guide in a new release.

PR #597, opened 2022-09-30
Update instancetype docs for version v1alpha2

lyarwood:
/area instancetype
/cc @akrejcir 
/cc @davidvossel 

This includes a new format for the data stored within the controllerrevision. This is updated in the versioning example.


davidvossel (commented):
/lgtm
/approve

PR #600, opened 2022-10-12
Disk sharing example PVC should use RWX access mode

aglitke:
The current example PVC uses a RWO access mode but if the point is to share this disk between multiple VMs then a RWX access mode would be more appropriate.

Signed-off-by: Adam Litke <alitke@redhat.com>


cwilkers:
Thanks!
/approve

PR #609, opened 2022-10-26
Update list of supported CRI implementations

rmohr:
We don't use docker anymore anywhere and it is deprecated. But we use containerd via kind and a vendors use it too.


maiqueb (approved):
Thank you for keeping us honest.

PR #611, opened 2022-10-26
Adds virtio-fs user guide

rverma-uc:
Signed-off-by: RahulV <rverma@upchain.com>

Adds `virtIO-FS` user guide.

### DIFF
--- a/docs/virtual_machines/disks_and_volumes.md
+++ b/docs/virtual_machines/disks_and_volumes.md
@@ -1402,3 +1402,69 @@ Test awesome shareable disks150+0 records in
 ```
 
 If you are using local devices or RWO PVCs, setting the affinity on the VMs that share the storage guarantees they will be scheduled on the same node. In the example, we set the affinity on the second VM using the label used on the first VM. If you are using shared storage with RWX PVCs, then the affinity rule is not necessary as the storage can be attached simultaneously on multiple nodes.
+
+## File Systems
+
+Currently, only `virtIO-FS` file system is supported.
+### virtIO-FS
+
+`virtIO-FS` makes a Persistent Volume Claim visible in the `kubevirt` VM.
+
+> **Note:** The **ExperimentalVirtiofsSupport** feature gate
+> [must be enabled](../operations/activating_feature_gates.md#how-to-activate-a-feature-gate)
+> to use this volume. This feature is available starting with KubeVirt v0.36.0.
+
+Example:
+
+```yaml
+    metadata:
+      name: testvmi-fs
+    apiVersion: kubevirt.io/v1
+    kind: VirtualMachineInstance
+    spec:
+      domain:
+        devices:
+          disks:
+          - disk:
+              bus: virtio
+            name: containerdisk
+          - disk:
+              bus: virtio
+            name: cloudinitdisk
+          filesystems:
+          - name: virtiofs-disk
+            virtiofs: {}
+        resources:
+          requests:
+            memory: 1024Mi
+      volumes:
+        - name: containerdisk
+          containerDisk:
+            image: quay.io/containerdisks/fedora:latest
+        - cloudInitNoCloud:
+          userData: |-
+            #cloud-config
+            password: fedora
+            chpasswd: { expire: False }
+          name: cloudinitdisk
+        - name: virtiofs-disk
+          persistentVolumeClaim:
+            claimName: mypvc
+```
+
+### File System Mounting
+
+`virtIO-FS` filesystem will not be available automatically once the VM launches. It needs to be mounted after the VM starts.
+
+This can be done in using startup script. See [cloudInitNoCloud](#cloudInitNoCloud) section for more details. Here are examples of how to mount it in a linux and windows VMs:
+
+- Linux Example
+
+```bash
+      sudo mkdir -p /mnt/disks/virtio
+      sudo mount -t virtiofs virtiofs-disk /mnt/disks/virtio
+```
+
+- Windows Example
+
+   See [this](https://virtio-fs.gitlab.io/howto-windows.html) guide for details on startup steps needed for Windows VMs.
--- a/docs/virtual_machines/disks_and_volumes.md
+++ b/docs/virtual_machines/disks_and_volumes.md
@@ -1442,10 +1442,10 @@ Example:
           containerDisk:
             image: quay.io/containerdisks/fedora:latest
         - cloudInitNoCloud:
-          userData: |-
-            #cloud-config
-            password: fedora
-            chpasswd: { expire: False }
+            userData: |-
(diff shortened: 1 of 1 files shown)


docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1402,3 +1402,58 @@ Test awesome shareable disks150+0 records in
...
+## File Systems
+
+Currently, only `virtIO-FS` file system is supported.
+### virtIO-FS
+
+`virtIO-FS` makes a Persistent Volume Claim visible in the `kubevirt` VM.
+
+> **Note:** The **ExperimentalVirtiofsSupport** feature gate
+> [must be enabled](../operations/activating_feature_gates.md#how-to-activate-a-feature-gate)
+> to use this volume. Available starting with KubeVirt v0.36.0.
```
chrisnegus, line 1415:
```suggestion
> to use this volume. This feature is available starting with KubeVirt v0.36.0.
```

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1402,3 +1402,58 @@ Test awesome shareable disks150+0 records in
...
+
+Example:
+
```
chrisnegus, line 1418: Add a line on its own of:
```
```yaml
```

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1402,3 +1402,58 @@ Test awesome shareable disks150+0 records in
...
+          requests:
+            memory: 128Mi
+      volumes:
+        - name: containerdisk
+          containerDisk:
+            image: quay.io/containerdisks/fedora:latest
+        - name: virtiofs-disk
+          persistentVolumeClaim:
+            claimName: mypvc
+
```
chrisnegus, line 1446: Add a line of its own with three backticks.

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1402,3 +1402,58 @@ Test awesome shareable disks150+0 records in
...
+### File System Mounting
+
+`virtIO-FS` filesystem will not be available automatically once the VM launches. It needs to be mounted after the VM starts.
+
+This can be done in using startup script. See [cloudInitNoCloud](#cloudInitNoCloud) section for more details. Here are examples of how to mount it in a linux and windows VMs:
+
+- Linux Example
+
```
chrisnegus, line 1454: Add three backticks and bash"

```
  ```bash
```

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1402,3 +1402,58 @@ Test awesome shareable disks150+0 records in
...
+      sudo mount /dev/virtiofs-disk /mnt/target
+
```
chrisnegus, line 1456: Add three backticks.

chrisnegus (commented):
Looks good. Just a couple of small things.

docs/virtual_machines/disks_and_volumes.md
rverma-uc: Fixed

docs/virtual_machines/disks_and_volumes.md
rverma-uc: Fixed

docs/virtual_machines/disks_and_volumes.md
rverma-uc: Fixed

docs/virtual_machines/disks_and_volumes.md
rverma-uc: Fixed

awels (commented):
/lgtm

lersek:
(Apologies if this is not the right place to ask the question below. Also adding @rwmjones due to potential general relevance of volume vs. VM disk association.)

Commit 5d597788133d adds three instances of the identifier (?) `virtiofs-disk`; one to `domain/devices/filesystems`, another to `volumes`, and finally one in the Linux guest `mount` command. I believe I understand the first and the last instances: the first one configures the virtiofs label on the host side, in the device model, and the last one enables the Linux guest kernel to find the same device based on label.

What role does the middle reference (the one under `volumes`) play? I'm generally unfamiliar with PVCs (persistent volume claims), but from other examples in the same manual, they seem to be Kubevirt's abstraction for "host-side backing files" or similar.

Now the reason I'm asking is this: the current API appears to bind together all three uses of of `virtiofs-disk`. Therefore, if for any reason you rename the *coupling* between the virtio-fs device model (= the first instance of the string in the commit) and the PVC (= the 2nd instance in the commit), that happens to change the label of the virtio filesystem, and exposes the guest as well to the change.

Is this scenario far fetched, or maybe known-and-tolerated? Would it be better if the virtio device/PVC coupling used one kind of coupling (separately from the virtiofs label), and the virtio device / guest Linux fs label coupling were isolated? That would mean the `domain/devices/filesystems` element would have to grow a further child knob, and the `volumes` entry would have to refer to that.

I'm not trying to propose useless over-engineering, it's just that on the raw QEMU command line, or in a libvirt domain XML, the host side config (what directory you share over virtiofs -- somewhat similarly to the PVC config here), and the guest side label, are strictly separate. Reorganizing PVC assignments in a YAML file should *perhaps* not result in a guest-visible change, but I'm unsure about the requirements in Kubevirt. Thanks.

awels:
With pods in general you will see 2 references to the same name.
1. In the volumes section of the pod spec. Here you name the volume, and you point it to a volume (PVC/EmptyDir/etc). This names the volume a specific name inside the context of the pod.
2. In the volumeMounts section of a container inside the pod spec. This is where you use the name you got in 1 at a particular mount point inside the container.

See https://kubernetes.io/docs/tasks/configure-pod-container/configure-volume-storage/ for an example.

For the virtio-fs example, we have the same 2 and an additional use inside the guest. We change the label inside the guest so it is easy to identify which volume inside the guest you use when mounting the volume inside the guest.

lersek:
... In effect I'm questioning Kubevirt commit 835249b0227e ("handle filesystem virtiofs devices", 2020-09-18); this part:

~~~
+               newFS.Target = &FilesystemTarget{
+                       Dir: fs.Name,
+               }
~~~

It corresponds to the domain XML fragment https://libvirt.org/kbase/virtiofs.html

~~~
      <target dir='mount_tag'/>
~~~

and hardwires `mount_tag` to the identifier that's internal to the YAML file. That's as if the `redis-storage` identifier from the YAML at https://kubernetes.io/docs/tasks/configure-pod-container/configure-volume-storage/ were exposed to the container. But now I understand this is not a problem for Kubevirt; thanks for the pointer.

PR #612, opened 2022-10-27
Document the support status of AppArmor

vasiliy-ul:
Fixes https://github.com/kubevirt/kubevirt/issues/7473

### DIFF
--- a/docs/operations/installation.md
+++ b/docs/operations/installation.md
@@ -27,6 +27,59 @@ KubeVirt is currently supported on the following container runtimes:
 Other container runtimes, which do not use virtualization features,
 should work too. However, they are not tested.
 
+### Integration with AppArmor
+
+In most of the scenarios, KubeVirt can run normally on systems with
+AppArmor. However, there are several known use cases that may require
+additional user interaction.
+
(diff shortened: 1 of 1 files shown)


docs/operations/installation.md
```diff
@@ -27,6 +27,59 @@ KubeVirt is currently supported on the following container runtimes:
 Other container runtimes, which do not use virtualization features,
 should work too. However, they are not tested.
 
+### Integration with AppArmor and SELinux
```
jean-edouard, line 30: nit: I don't see anything about SELinux in this section
```suggestion
### Integration with AppArmor
```

jean-edouard (commented):
/approve
Thank you!

docs/operations/installation.md
vasiliy-ul: Initially, I wanted to have a section about MACs in general. But then realized that there is basically nothing else to say about SELinux apart that KubeVirt works fine with it. So this is currently just emphasized in the first sentence:
> "In most of the scenarios, KubeVirt can run normally on systems with AppArmor or **SELinux**."

Still not sure if it makes sense to mention SELinux at all. I will probably drop it then, unless there are other thoughts on that...

PR #615, opened 2022-11-07
virtiofs, hostpath: Add an example for using hostpath in VMs.

jcanocan:
Add an example explaining how to share node directories with VMs using virtiofs. This example is based in this [discussion](https://groups.google.com/g/kubevirt-dev/c/zTnXrjYDDcc).

### DIFF
--- a/docs/virtual_machines/disks_and_volumes.md
+++ b/docs/virtual_machines/disks_and_volumes.md
@@ -1403,68 +1403,217 @@ Test awesome shareable disks150+0 records in
 
 If you are using local devices or RWO PVCs, setting the affinity on the VMs that share the storage guarantees they will be scheduled on the same node. In the example, we set the affinity on the second VM using the label used on the first VM. If you are using shared storage with RWX PVCs, then the affinity rule is not necessary as the storage can be attached simultaneously on multiple nodes.
 
-## File Systems
+## Sharing Directories with VMs
 
-Currently, only `virtIO-FS` file system is supported.
-### virtIO-FS
-
-`virtIO-FS` makes a Persistent Volume Claim visible in the `kubevirt` VM.
+`Virtiofs` allows to make visible external filesystems to `KubeVirt` VMs.
+`Virtiofs` is a shared file system that lets VMs access a directory tree on the host.
+Further details can be found at [Official Virtiofs Site](https://virtio-fs.gitlab.io/).
 
 > **Note:** The **ExperimentalVirtiofsSupport** feature gate
 > [must be enabled](../operations/activating_feature_gates.md#how-to-activate-a-feature-gate)
-> to use this volume. This feature is available starting with KubeVirt v0.36.0.
+> to share directories with VMs. This feature is available starting with KubeVirt v0.36.0.
 
+### Sharing Persistent Volume Claims
+#### Cluster Configuration
+
+We need to create a new VM definition including the `spec.devices.disk.filesystems.virtiofs` and a PVC.
 Example:
 
 ```yaml
-    metadata:
-      name: testvmi-fs
-    apiVersion: kubevirt.io/v1
-    kind: VirtualMachineInstance
-    spec:
-      domain:
-        devices:
-          disks:
-          - disk:
-              bus: virtio
-            name: containerdisk
-          - disk:
-              bus: virtio
-            name: cloudinitdisk
-          filesystems:
-          - name: virtiofs-disk
-            virtiofs: {}
-        resources:
-          requests:
-            memory: 1024Mi
-      volumes:
-        - name: containerdisk
-          containerDisk:
-            image: quay.io/containerdisks/fedora:latest
-        - cloudInitNoCloud:
-            userData: |-
-              #cloud-config
-              password: fedora
-              chpasswd: { expire: False }
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: testvmi-fs
+spec:
+  domain:
+    devices:
+      disks:
+        - disk:
+            bus: virtio
+          name: containerdisk
+        - disk:
+            bus: virtio
           name: cloudinitdisk
+      filesystems:
         - name: virtiofs-disk
-          persistentVolumeClaim:
-            claimName: mypvc
+          virtiofs: {}
+    resources:
+      requests:
+        memory: 1024Mi
+  volumes:
+    - name: containerdisk
+      containerDisk:
+        image: quay.io/containerdisks/fedora:latest
+    - cloudInitNoCloud:
+        userData: |-
+          #cloud-config
+          password: fedora
+          chpasswd: { expire: False }
+      name: cloudinitdisk
+    - name: virtiofs-disk
+      persistentVolumeClaim:
+        claimName: mypvc
 ```
 
-### File System Mounting
-
-`virtIO-FS` filesystem will not be available automatically once the VM launches. It needs to be mounted after the VM starts.
-
-This can be done in using startup script. See [cloudInitNoCloud](#cloudInitNoCloud) section for more details. Here are examples of how to mount it in a linux and windows VMs:
+#### Configuration Inside the VM
+The following configuration can be done in using startup script. See [cloudInitNoCloud](#cloudinitnocloud) section for 
+more details. 
+However, we can do it manually by logging in to the VM and mounting it. 
+Here are examples of how to mount it in a linux and windows VMs:
 
 - Linux Example
 
 ```bash
-      sudo mkdir -p /mnt/disks/virtio
-      sudo mount -t virtiofs virtiofs-disk /mnt/disks/virtio
+$ sudo mkdir -p /mnt/disks/virtio
+$ sudo mount -t virtiofs virtiofs-disk /mnt/disks/virtio
 ```
 
 - Windows Example
 
-   See [this](https://virtio-fs.gitlab.io/howto-windows.html) guide for details on startup steps needed for Windows VMs.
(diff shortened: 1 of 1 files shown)


docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1468,3 +1468,136 @@ This can be done in using startup script. See [cloudInitNoCloud](#cloudInitNoClo
...
+              user: fedora
+          name: cloudinitdisk
+        - name: vm-hostpath
+          persistentVolumeClaim:
+            claimName: hostpath-claim
+```
+
+
+#### Hostpath Mounting
+To mount the hostpath in the VM, we have to access the node for setting the proper SELinux context to the shared directory:
```
dhiller, line 1593: I can't follow why we need to do this after we create the PV and PVC? Wouldn't this be something we would need to prepare before?

dhiller (changes requested):

docs/virtual_machines/disks_and_volumes.md
jcanocan: The system does not set the SELinux context by itself and as far as I know, this is an intended behavior. @xpivarc Can you provide further details on this please?

germag: Could be possible to start with setting up the directory labels?, I mean, something like (not intended to be copied verbatim :) just a rough idea):
"
Let's assume we want to share the node directory `/mnt/share`, first we must check the security label of  `/mnt/share`, otherwise the VM will not be able to access that directory:
``` example of to check sec label ``
if the sec label isn't present or is not `container_file_t` we need to label it as:
``` sudo chcon ... ```
"

So, I think a structure like this is a bit clearer:
- Checking/Setting up the shared directory
- all the yamls
- Hostpath mounting (with instructions from inside the VM only)

**Comment by dominikholler** [comment]
@alicefr Is the change fine from your point of view?

**Comment by alicefr** [comment]
Why are showing an example with hostpath and not a more generic one with simply using pvcs?

File: docs/virtual_machines/disks_and_volumes.md (hunk 1 above)
Comment (alicefr) [reply docs/virtual_machines/disks_and_volumes.md:1593]: I need to admit I have never tried with hostpath, but with a proper storage class, the selinux label should be handled directly from the CSI plugin. Don't we want to illustrate a more generic example with PVCs?

File: docs/virtual_machines/disks_and_volumes.md [hunk 2]
```diff
@@ -1468,3 +1468,136 @@ This can be done in using startup script. See [cloudInitNoCloud](#cloudInitNoClo
...
+
+```shell
+sudo chcon -t container_file_t /tmp/data
```
alicefr, line 1596: Where should this command be run? Directly on host or virt-launcher?

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1468,3 +1468,136 @@ This can be done in using startup script. See [cloudInitNoCloud](#cloudInitNoClo
 - Windows Example
 
    See [this](https://virtio-fs.gitlab.io/howto-windows.html) guide for details on startup steps needed for Windows VMs.
+
+### Sharing Node Directories with VMs
+
+`Virtiofs` allows to share node directories with VMs as [hostpaths](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath).
```
alicefr, line 1474: IMO, this sentence is a bit misleading. Virtiofs allows sharing directories directly with the guest, not only host paths.  The directory is provided through a PVC, and this can be backed by a hostpath

docs/virtual_machines/disks_and_volumes.md
xpivarc: > The system does not set the SELinux context by itself and as far as I know, this is an intended behavior. @xpivarc Can you provide further details on this please?

Yes, but at the same time, it depends on what approach you choose. See the above comments.

alicefr: @xpivarc yes, I agree with you.
I understand that this initial example is targeting hostpath specifically. But on the other hand, we don't have any other good examples with virtiofs, and I think it is important to explain to the users that they have other (better) solutions. 
So, as @xpivarc already said, I think it is important to mention the other cases

jcanocan: Thanks for your comments. Regarding @germag suggestion, I agree and I have changed the structure as described. I think this structural change could solve some of your concerns.
About the generic example with PVCs, do you mean an example like [this](https://github.com/kubevirt/user-guide/blob/main/docs/virtual_machines/disks_and_volumes.md?plain=1#L1406-L1470)? I have reworked a bit the existing [example](https://github.com/kubevirt/user-guide/blob/main/docs/virtual_machines/disks_and_volumes.md?plain=1#L1406-L1470) and mine to reflect your suggestions. Furthermore, I have added a few lines explaining that the hostpath method is discouraged.
Additionally, I guess that the third method mentioned is using a proper storage class. However, I have never tried this approach and I'm not sure how to do it. If you have some hints in this regard please, let me know. On the other hand, because I need some time to learn and test it, I will post it in a separate PR.

docs/virtual_machines/disks_and_volumes.md
jcanocan: I agree. I have changed it.

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1403,68 +1403,217 @@ Test awesome shareable disks150+0 records in
...
+      name: cloudinitdisk
+    - name: virtiofs-disk
+      persistentVolumeClaim:
+        claimName: mypvc
+```
+
+#### Configuration Inside the VM
+The following configuration can be done in using startup script. See [cloudInitNoCloud](#cloudinitnocloud) section for 
+more details. 
+However, we can do it manually by login inside the VM and mounting it.
```
dhiller, line 1461: nit
```suggestion
However, we can do it manually by logging into the VM and mounting it. 
```

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1403,68 +1403,217 @@ Test awesome shareable disks150+0 records in
...
+
+- Windows Example
+
+  See [this](https://virtio-fs.gitlab.io/howto-windows.html) guide for details on startup steps needed for Windows VMs.
+
+### Sharing Node Directories
+
+It is allowed using [hostpaths](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath).
+The following configuration example is shown for illustrative purposes.
+However, the [PVCs](#sharing-persistent-volume-claims) method is preferred since hostpath are generally discouraged for
```
dhiller, line 1479: nit
```suggestion
However, the [PVCs](#sharing-persistent-volume-claims) method is preferred since using hostpath is generally discouraged for
```

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1403,68 +1403,217 @@ Test awesome shareable disks150+0 records in
...
+security reasons.
+
+
+#### Configuration Inside the Node
+
+We need to log in the node which we want to share the directory with the VMs, to create the shared directory (if it is not
```
dhiller, line 1485: nit
```suggestion
We need to log into the node of which we want to share the directory with the VMs, to create the shared directory (if it is not
```

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1403,68 +1403,217 @@ Test awesome shareable disks150+0 records in
...
-- Linux Example
+#### Configuration Inside the VM
 
-```bash
-      sudo mkdir -p /mnt/disks/virtio
-      sudo mount -t virtiofs virtiofs-disk /mnt/disks/virtio
-```
-
-- Windows Example
+We need to access the VM and mount the shared directory:
```
dhiller, line 1615: nit
```suggestion
We need to log into the VM and mount the shared directory:
```

dhiller (approved):
Thanks @jcanocan for this great work, I can follow now :)

I've found some minor nits, which you might disregard, since I am no native speaker.

Maybe @stu-gott or @aburdenthehand could take a look also?

docs/virtual_machines/disks_and_volumes.md
aburdenthehand: To nit your nit, it should be "...by logging in to the VM..."
('Logging in' or 'log in' is a phrasal verb and 'to' is the preposition that connects it to the object.)

docs/virtual_machines/disks_and_volumes.md
aburdenthehand: This is quite a complicated sentence so I'd like to suggest the following, starting with the objective and followed by the steps needed:
"To share the directory with the VMs, we need to log in to the node, create the shared directory (if it does not already exist), and set the proper SELinux context label `container_file_t` to the shared directory."
Let me know if I misunderstand.

docs/virtual_machines/disks_and_volumes.md
aburdenthehand: "We need to log in to the VM..."

docs/virtual_machines/disks_and_volumes.md
jcanocan: Thanks for the clarification. Done.

docs/virtual_machines/disks_and_volumes.md
jcanocan: Done

docs/virtual_machines/disks_and_volumes.md
jcanocan: You have understand it perfectly fine. Thanks for the rework :)

alicefr:
Looks good to me! 
/lgtm

PR #619, opened 2022-11-09
passt: Remove note about extra memory when using passt

rhrazdil:
When configuring passt binding with no explicit ports to forward, the memory overhead needed is added automatically, users no longer need to provide the VM with extra memory.

See https://github.com/kubevirt/kubevirt/pull/8716

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

### DIFF
--- a/docs/virtual_machines/interfaces_and_networks.md
+++ b/docs/virtual_machines/interfaces_and_networks.md
@@ -649,7 +649,6 @@ When no ports are explicitly specified, all ports are forwarded, leading to memo
 
 #### Temporary restrictions: 
 1. `passt` currently only supported as primary network and doesn't allow extra multus networks to be configured on the VM.
-2. A VM with passt binding cannot bind to low ports (less than 1024).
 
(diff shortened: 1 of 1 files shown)


docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -643,7 +643,6 @@ sysctl -w net.core.wmem_max = 33554432
 ```
 sysctl -w fs.file-max = 9223372036854775807
 ```
-3. The `vmi.Spec.Domain.Resources.Requests.Memory` of a passt VM with no explicit ports should be at least 2048M.
```
AlonaKaplan, line 646: Please add a warning that a passt VM with no explicit ports has an increased memory overhead.

rhrazdil: Added.
Also I've noticed there is still mention of a limitation saying binding to low ports is not possible, I've removed it as well as it's no longer the case.

PR #625, opened 2022-11-30
Removing ornamental which statement due to missing in container

cwilkers:
The `which` binary is not installed in the container for back-end updates, and this is causing make to fail.

Signed-off-by: cwilkers <cwilkers@redhat.com>


cwilkers:
@dhiller if you have a chance, take a look

PR #626, opened 2022-12-05
Add the need to stop VM before restore

ShellyKa13:
To prevent confusion add explicitly the need to stop VM before being able to restore it


davidvossel (approved):
/lgtm
/approve

PR #633, opened 2023-01-10
Switch to Material theme

phoracek:
This theme has a more modern look that can be synced with the colors of KubeVirt, it has a clear documentation and is more flexible when it comes to dividing content into sections.

In follow-up patches, the content will be reorganized to be easier to navigate.

Note that this change depends on: https://github.com/kubevirt/project-infra/pull/2551


aburdenthehand (approved):
Thanks for the PR, Petr! This looks really great. 
This is a fairly large change to the docs so I've tried to compare this new theme with the old one, looking at multiple aspects. In almost every way this is an improvement. 

Navigation
+ Both themes display up to 4 levels of ToC depth. Material theme separates first level from the other levels into a separate side panel, making it easier to view more of the ToC at any given time. This also means it can automatically explodes the additional layers, rather than having to manually open them.
+ Clicking a level 1 heading (left panel) in material centres the ToC on that heading. This is preferable to the current theme which reloads the page to the top of the ToC. 

Display
+ New theme uses larger proportion of the screen and a larger font makes it easier to read.
- New theme has less distinction between the chapter headings (Operations, etc) and the level 1 headings in the ToC
- Monospace/code text within para's in the new theme don't stand out as much in the new theme. 
(It looks like these two last points can be improved/fixed with either config changes or extensions) 

Search
Compared the search function between the two themes and the material theme offers the following improvements:
+ results are in a pop-up, so you can search for a term without navigating away from your current page. (If you hit enter, it loads the first search result).
+ If multiple hits are on the same page, it does not display all of them but collapses them into a 'X more on this page'. IMO this makes the results much easier to navigate.
+ Shows the para containing the search term and highlights it. (Old theme displays the first 200 characters of the section that contains the search term.)   

In most cases the two searches return the same results. The exception is that the material search does not recognise search terms if they appear within parenthesis, and presumably when not prefixed by whitespace. Looks like this can be fixed by adding additional separators: https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/#+search.separator
We can do this is in a subsequent PR.

phoracek:
Thanks for the review! I believe that the three last commits should address your concerns.

aburdenthehand:
/lgtm
Headings and code text are more discernible and search has been improved. 
Really nice!

fabiand:
/approve

Awesome, thank you for the … dedication.

PR #639, opened 2023-01-31
Adds ability to switch between dark and light modes

dharmit:
Signed-off-by: Dharmit Shah <shahdharmit@gmail.com>

I've used the configuration provided [in the official docs](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/?h=palette#color-palette-toggle).

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -33,8 +33,18 @@ theme:
   logo: assets/KubeVirt_icon.png
   favicon: assets/favicon32x32.png
   palette:
-    primary: teal
-    accent: teal
+    # Palette toggle for light mode
+    - scheme: default
+      toggle:
+        icon: material/weather-night
+        name: Switch to dark mode
+
+    # Palette toggle for dark mode
+    - scheme: slate
+      toggle:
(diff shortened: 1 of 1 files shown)


cwilkers:
The preview looks really good. I checked a couple pages with images (like the architecture overview), and they seem fine as well. Thanks for this!
I'll put in my 
/lgtm
And see if any others would like to comment before final approval.

aburdenthehand:
/lgtm
I've gone through a variety of pages and have not seen any impact on how light mode works vs our current docs.
I see no downside to offering dark mode as an option for users, and personally think it's a really great addition. 
We might want to slightly change the palette for dark mode to make hyperlinks a bit easier to read. But that can happen in a separate PR. Similarly, some of the diagrams which have black lines or text with no background get lost in dark mode, but it's easy for user to toggle for those kinds of situations. 
@dharmit Thank you for submitting this!

dharmit:
@cwilkers @aburdenthehand both of you gave an lgtm on this. Can one of you approve it instead so that this goes in?

PR #641, opened 2023-02-01
Add prerequisites for the guest to get the correct MTU

AlonaKaplan:
Add prerequisites for the guest to get the correct MTU


EdDev (approved):
Looks good.
It will be even better if we provide the reader a version of something in order to identify this is indeed the problem.

PR #643, opened 2023-02-07
Windows guest non virtio interfaces don't get the correct MTU

AlonaKaplan:
Update the networking doc.


EdDev (approved):
Consider a newline in between this section and the prev.

phoracek:
That note is becoming a a junkyard for MTU comments. I agree with it Edy, let's create a new paragraph, starting with `Note (Windows):` may be helpful

phoracek (approved):
/approve

Thanks

PR #653, opened 2023-03-10
Remove Web Console documentation

phoracek:
The page is out of date. Keeping it in sync with the development of the Web Console would require a lot of attention. Also, the UI should be self-explanatory.

The second reason for this change is, that this specific UI implementation is connected with a specific vendor. Removing it would keep the user-guide neutral.


aburdenthehand:
/lgtm
I agree on all points. And I can't see any additional links to this content in the rest of the user-guide. Thanks Petr!

PR #654, opened 2023-03-13
Bug - Fixed web console wizard tables

### LINKED ISSUE #594: Web console wizard tables malformatted
Looks like the two first tables in the web console have contents in the wrong cells:
https://kubevirt.io/user-guide/web_console/#virtual-machine-template-wizard-fields
https://kubevirt.io/user-guide/web_console/#virtual-machine-wizard-fields

To fix, I think we need to change the following:
1. 'PXE', 'URL', and 'Container' should all be in the 'Parameter' column of 'Provision Source', with associated descriptions in the third column. 
2. 'generic' and 'highperformance' should be in the 'Parameter' column of ' Workload Profile', with associated description in the third column.
3. Specific only to the Virtual Machine Template Wizard Fields table: I think we can remove the first 'Provision Source' row (seems to be confused with the second 'Provision Source' row) and the 'Template' row (as I don't think you can create a template using a template).

File to update: web_console.md

Additional quests: Are we able to make the second column (Parameter) thinner? It is currently pushing the Description column off-screen. 
Are these fields still valid? I have a feeling these are from a much earlier version of KubeVirt.

adithyaakrishna:
**Description:**
This PR Fixes #594

### DIFF
--- a/docs/web_console.md
+++ b/docs/web_console.md
@@ -553,12 +553,12 @@ Details** screen.
 | Description || Optional description field |
 | Template || Template from which to create the virtual machine<br><br>* Selecting a template will automatically fill other fields |
 | Provision Source | PXE ||
-| URL | Provision virtual machine from an image available from an *HTTP* or *S3* endpoint ||
-| Container | Provision virtual machine from a bootable operating system container located in a registry accessible from the cluster<br><br>* Example: `_kubevirt/cirros-registry-disk-demo_` ||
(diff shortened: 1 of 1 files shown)


phoracek:
Hello @adithyaakrishna, thanks for posting this. Unless anybody protests against https://github.com/kubevirt/user-guide/pull/653, the Web Console page will be removed from our user-guide, so this PR will become redundant

adithyaakrishna:
Hi @phoracek 
Makes sense, I totally understand that, please feel free to close this PR in-case the above PR gets merged :)

PR #665, opened 2023-03-27
Add documentation for vmexport manifests

awels:
Explain the new 'manifests' field with the associated links for the manifest.

Explain the virtctl support for manifest so that it is easier to copy a VM to another cluster. Added example for this work flow.

### DIFF
--- a/docs/operations/export_api.md
+++ b/docs/operations/export_api.md
@@ -202,9 +202,9 @@ status:
 ```
 
 #### Manifests
-The VirtualMachine manifests can be retrieved by accessing the `manifests` in the VirtualMachineExport status. The `all` type will return the VirtualMachine manifest, any DataVolumes, and a configMap that contains the public CA certificate of the Ingress/Route of the external URL, or the CA of the export server of the internal URL. The `auth-header-secret` will be a secret that contains a Containerized Data Import (CDI) compatible header. This header contains a text version of the export token.
+The VirtualMachine manifests can be retrieved by accessing the `manifests` in the VirtualMachineExport status. The `all` type will return the VirtualMachine manifest, any DataVolumes, and a configMap that contains the public CA certificate of the Ingress/Route of the external URL, or the CA of the export server of the internal URL. The `auth-header-secret` will be a secret that contains a Containerized Data Importer (CDI) compatible header. This header contains a text version of the export token.
 
-Both internal and external links will contain a `manifests` field. If there are no external links, then there will not be any external manifests either.
+Both internal and external links will contain a `manifests` field. If there are no external links, then there will not be any external manifests either. The virtualMachine `manifests` field is only available if the source is a `VirtualMachine` or `VirtualMachineSnapshot`. Exporting a `PersistentVolumeClaim` will not generate a Virtual Machine manifest.
 
 ```yaml
 apiVersion: export.kubevirt.io/v1alpha1
@@ -512,14 +512,14 @@ Now we are ready to import this disk into the target cluster. In order for CDI t
 
 virtctl provides an additional argument to the download command called `--manifest` that will retrieve the appropriate information from the export server, and either save it to a file with the `--output` argument or write to standard out. By default this output will not contain the header secret as it contains the token in plaintext. To get the header secret you specify the `--include-secret` argument. The default output format is `yaml` but it is possible to get `json` output as well.
 
-Assuming there is a running Virtual Machine export called `example-export` and the same namespace exists in the target cluster. The name of the kubeconfig of the target cluster is named `kubeconfig-target`, to clone the vm into the target cluster run the following commands:
+Assuming there is a running VirtualMachineExport called `example-export` and the same namespace exists in the target cluster. The name of the kubeconfig of the target cluster is named `kubeconfig-target`, to clone the vm into the target cluster run the following commands:
 
 ```bash
 $ virtctl vmexport download example-export --manifest --include-secret --output=import.yaml
 $ kubectl apply -f import.yaml --kubeconfig=kubeconfig-target
 ```
 
-The first command generates the yaml and writes it to `import.yaml`. The second command applies the yaml to the target cluster. It is possible to combine the two commands by not writing to a file, and reading from standard in in the second command. Use this option if the export token should not be written to a file anywhere. This will create the VM in the target cluster, and provides CDI in the target cluster with everything required to import the disk images into the target cluster.
+The first command generates the yaml and writes it to `import.yaml`. The second command applies the generated yaml to the target cluster. It is possible to combine the two commands writing to standard `out` with the first command, and piping it into the second command. Use this option if the export token should not be written to a file anywhere. This will create the VM in the target cluster, and provides CDI in the target cluster with everything required to import the disk images.
 
 After the import completes you should be able to start the VM in the target cluster.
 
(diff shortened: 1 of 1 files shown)


docs/operations/export_api.md
```diff
@@ -200,6 +200,55 @@ status:
   phase: Ready
   serviceName: virt-export-example-export
 ```
+
+#### Manifests
+The VirtualMachine manifests can be retrieved by accessing the `manifests` in the VirtualMachineExport status. The `all` type will return the VirtualMachine manifest, any DataVolumes, and a configMap that contains the public CA certificate of the Ingress/Route of the external URL, or the CA of the export server of the internal URL. The `auth-header-secret` will be a secret that contains a Containerized Data Import (CDI) compatible header. This header contains a text version of the export token.
+
+Both internal and external links will contain a `manifests` field. If there are no external links, then there will not be any external manifests either.
+
```
jpeimer, line 208: Should we also add something like: 
VirtualMachine manifest will be present only if VirtualMachineExport was created for VirtualMachine or VirtualMachineSnapshot. Export of PersistentVolumeClaim will not generate a VirtualMachine manifest.

docs/operations/export_api.md
```diff
@@ -455,80 +504,22 @@ status:
...
-data:
-  ca.pem: |
-    -----BEGIN CERTIFICATE-----
-    ...
-    -----END CERTIFICATE-----
-```
+virtctl provides an additional argument to the download command called `--manifest` that will retrieve the appropriate information from the export server, and either save it to a file with the `--output` argument or write to standard out. By default this output will not contain the header secret as it contains the token in plaintext. To get the header secret you specify the `--include-secret` argument. The default output format is `yaml` but it is possible to get `json` output as well.
 
-Next create a secret in the same namespace on the target cluster. Note: the token value will be a header that is passed to the server by CDI.
+Assuming there is a running Virtual Machine export called `example-export` and the same namespace exists in the target cluster. The name of the kubeconfig of the target cluster is named `kubeconfig-target`, to clone the vm into the target cluster run the following commands:
```
jpeimer, line 515: VirtualMachineExport called `example-export`

jpeimer: > and the same namespace exists in the target cluster. The name of the kubeconfig of the target cluster is named `kubeconfig-target`

Do we really need this part?

docs/operations/export_api.md
```diff
@@ -455,80 +504,22 @@ status:
...
-```yaml
-apiVersion: v1
-kind: Secret
-metadata:
-  name: secret-headers
-stringData:
-  token: "x-kubevirt-export-token:1234567890ab"
+```bash
+$ virtctl vmexport download example-export --manifest --include-secret --output=import.yaml
+$ kubectl apply -f import.yaml --kubeconfig=kubeconfig-target
```
jpeimer, line 519: > --kubeconfig=kubeconfig-target

If the kubeconfig was exported, it should work without this, right?

docs/operations/export_api.md
```diff
@@ -455,80 +504,22 @@ status:
...
-        resources:
-          requests:
-            memory: 2Gi
-      terminationGracePeriodSeconds: 0
-      volumes:
-      - dataVolume:
-          name: example-dv
-        name: datavolumedisk1
-```
+The first command generates the yaml and writes it to `import.yaml`. The second command applies the yaml to the target cluster. It is possible to combine the two commands by not writing to a file, and reading from standard in in the second command. Use this option if the export token should not be written to a file anywhere. This will create the VM in the target cluster, and provides CDI in the target cluster with everything required to import the disk images into the target cluster.
```
jpeimer, line 522: > standard in in the second command

standard input?

jpeimer: > This will create the VM in the target cluster, and provides CDI in the target cluster with everything required to import the disk images into the target cluster.

This will create the VM in the target cluster, and will provide CDI in the target cluster with everything required to import the disk images.

jpeimer (changes requested):
Thanks a lot for the doc! 
Just a few suggestions:

docs/operations/export_api.md
```diff
@@ -200,6 +200,55 @@ status:
   phase: Ready
   serviceName: virt-export-example-export
 ```
+
+#### Manifests
+The VirtualMachine manifests can be retrieved by accessing the `manifests` in the VirtualMachineExport status. The `all` type will return the VirtualMachine manifest, any DataVolumes, and a configMap that contains the public CA certificate of the Ingress/Route of the external URL, or the CA of the export server of the internal URL. The `auth-header-secret` will be a secret that contains a Containerized Data Import (CDI) compatible header. This header contains a text version of the export token.
```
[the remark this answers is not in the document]
awels: Fixed

docs/operations/export_api.md
awels: Done

docs/operations/export_api.md
awels: Yes we do, there are 2 kubeconfigs at play here, the `source` which is the exported KUBECONFIG, and the target, which has to be specified.

docs/operations/export_api.md
awels: No, we are apply the generated manifest to the target cluster which has a different kubeconfig than the source cluster which we used for the first command.

docs/operations/export_api.md
awels: reworded it a little bit to explain the intent better.

docs/operations/export_api.md
jpeimer: Okay, it's needed if we are accessing the target cluster from the same tab

docs/operations/export_api.md
```diff
@@ -455,80 +504,22 @@ status:
...
-        resources:
-          requests:
-            memory: 2Gi
-      terminationGracePeriodSeconds: 0
-      volumes:
-      - dataVolume:
-          name: example-dv
-        name: datavolumedisk1
-```
+The first command generates the yaml and writes it to `import.yaml`. The second command applies the generated yaml to the target cluster. It is possible to combine the two commands writing to standard `out` with the first command, and piping it into the second command. Use this option if the export token should not be written to a file anywhere. This will create the VM in the target cluster, and provides CDI in the target cluster with everything required to import the disk images.
```
jpeimer, line 523: There are more good examples in your PR, maybe we should add them?

Examples :
get manifest for existing export, that will NOT include header secret
```bash
$ virtctl vmexport download export --manifest
```
create vm export for vm example, and get the manifest, it will NOT include header secret
```bash
$ virtctl vmexport download export --manifest --vm=example
```
create vm export for vm snapshot example, and get the manifest, it will NOT include header secret
```bash
$ virtctl vmexport download export --manifest --snap=example
```
get manifest for existing export, that will include the header secret
```bash
$ virtctl vmexport download export --manifest --include-secret
```
get manifest for existing export, that will NOT include the header secret, in json format
```bash
$ virtctl vmexport download export --manifest --manifest-output-format=json
```
get manifest for existing export, that will include the header secret writing to a file
```bash
$ virtctl vmexport download export --manifest --include-secret --output=manifest.yaml
```

jpeimer (commented):
Thanks a lot for the update! 
Looks good, just a suggestion to add more examples

docs/operations/export_api.md
awels: The `virtctl export` should have examples of all these combinations. it would make sense to keep them there so if they change the documentation is not out of date.

jpeimer: Ok, thanks

aglitke:
Looks good to me as well.

PR #670, opened 2023-04-05
Typo

jean-edouard:


docs/operations/snapshot_restore_api.md
```diff
@@ -61,9 +61,9 @@ kubectl wait vmsnapshot snap-larry --for condition=Ready
 ```
 
 You can check the vmSnapshot phase in the vmSnapshot status. It can be one of the following:
-* InProgress
-* Succeeded
-* Failed.
+- InProgress
+- Succeeded
+- Failed.
```
cwilkers, line 66:
```suggestion

- InProgress
- Succeeded
- Failed.
```

The parser needs a blank newline after the previous section to offset the list.

cwilkers (changes requested):
The list still doesn't render (this is one of those cases where the local preview render helps a lot) it turns out the magic is the newline above the list. You could even have left the asterisks in (but the dashes match other pages better)

cwilkers:
The render checks out now, thanks!
/lgtm
/approve

Issue #678, opened 2023-04-26
Interfaces and Networks documentation uses example that is known to not work. 


radmacd:
Linked documentation uses an example with a macvlan CNI + bridge interface type combination that is known to not work.

https://github.com/kubevirt/user-guide/blob/b854df59af906d10c17395749469dde0573bcab9/docs/virtual_machines/interfaces_and_networks.md?plain=1#L147

/cc @maiqueb

PR #679, opened 2023-05-03
network: remove macvlan examples

maiqueb:
The provided examples on the `interfaces and networks` section are using the macvlan plugin, which we later on list as something that doesn't work.

This commit provides a functional example to the user instead, which will provide a better experience to the user (it works, and it is not listed as not working later on).

Fixes: #678


phoracek (approved):
/lgtm
/approve

Thanks!

PR #681, opened 2023-05-10
Improve cloud-init part of creating VMs

0xFelix:
Instead of adding an additional user with cloud init the example now shows how an authorized key can be added to the default user.


lyarwood:
> Instead of adding an additional user with cloud init the example now shows how an authorized key can be added to the default user.

@0xFelix  Would you mind updating the instance type example as well in this PR?

lyarwood (changes requested):
Instead of adding an additional user with cloud init the example now shows how an authorized key can be added to the default user.

@0xFelix Would you mind updating the instance type example as well in this PR?

0xFelix:
@lyarwood Done

cwilkers:
Looks very straight-forward to me. The only reservation I have is in how would a user necessarily know which is the default user for an arbitrary cloud image? If it's "Read the documentation for the cloud image" then that may be worth mentioning at least in the creating_vms doc.

PR #684, opened 2023-05-19
POC: Using index.md to simplify restructuring

aburdenthehand:
Following on from this [thread](https://groups.google.com/g/kubevirt-dev/c/HgvuIUef4SU) on the kubevirt-dev mailing list, it seemed easier to prototype on a branch with a preview build than a googledoc. 

Restructuring our current content into multiple categories to help users find the content they're looking for. 
Moving docs files between folders became laborious so perhaps it will be easier going forward to have all content in the docs/ folder. We can then use the index.md to restructure as needed.

Also added category description to the 'Welcome' splash page. These would eventually be better as links.


cwilkers:
Looks like interceding PRs have put this one in conflict with main. Do you want to fix and get this to a mergeable state?

aburdenthehand:
@cwilkers This really was just a POC but if folks are happy with this kind of structure then I can fix it up. Each ~chapter is in rough alphabetical order so I'll need to re-order them.

PR #685, opened 2023-06-07
hotplug-interfaces: Use VirtualMachine in the example

EdDev:
The user facing API is the VirtualMachine, it is less common to work with the back-end VirtualMachineInstance object.

Therefore, the example has been adjusted to describe a VM.

In addition, it has been simplified a bit by removing the cloud-init disk/volume.


orelmisan (approved):
Thank you @EdDev

PR #689, opened 2023-06-21
hotplug interface: allowed only on VMs

AlonaKaplan:
The --persist option was removed. Any interface hotplug will be persisted.

### DIFF
--- a/docs/operations/hotplug_interfaces.md
+++ b/docs/operations/hotplug_interfaces.md
@@ -1,4 +1,7 @@
 # Hotplug Network Interfaces
+
+**Warning**: The feature is in alpha stage and its API may be changed in the future.
+
 KubeVirt now supports hotplugging network interfaces into a running Virtual
 Machine (VM). Hotplug is only supported for interfaces using the
 `virtio` model connected through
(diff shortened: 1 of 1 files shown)


AlonaKaplan:
/hold
until https://github.com/kubevirt/kubevirt/pull/9954 is merged.

docs/operations/hotplug_interfaces.md
```diff
@@ -81,6 +81,7 @@ to the following snippet:
 ```bash
 virtctl addinterface vm-fedora --network-attachment-definition-name new-fancy-net --name dyniface1
 ```
+This will add the interface to the running VMI object, and also mutate the VM spec template.
```
EdDev, line 84: Maybe it is worth focusing on the VM and only then adding that "as a consequence", the running VMI is updated.

EdDev (changes requested):
Can you please also add at the head of the file that this feature is introduced as `alpha1` for tech-preview? Warning that changes to the API may occur per feedback from usage and other requirements.

docs/operations/hotplug_interfaces.md
AlonaKaplan: Changed to- "This will add the interface and network to the VM spec template and to the running VMI object.
".

EdDev (approved):

dhiller (approved):
/approve

PR #691, opened 2023-06-26
Update disks_and_volumes.md (typo)

shamil2:
Fixed typo:
claimName instead of clainName


cwilkers:
Nice catch, thanks for the fix!
/ok-to-test
/approve
/lgtm

PR #692, opened 2023-06-27
introduce VM and VMI workload types

maiqueb:
This new page attempts to provide the users information about when to use a VM or a VMI.

It highlights the differences between those user facing API objects.

### DIFF
--- a/docs/virtual_machines/vms_and_vmis.md
+++ b/docs/virtual_machines/vms_and_vmis.md
@@ -1,0 +1,41 @@
+# Virtual Machine types
+There are two main virtual machine user facing API objects: a `VirtualMachine`,
+and a `VirtualMachineInstance`.
+
+## Virtual Machine Instance
+A
+[VirtualMachineInstance](https://kubevirt.io/user-guide/virtual_machines/virtual_machine_instances/)
+represents a single **running** virtual machine.
+
+It is encapsulated in a pod.
+
+When used without a controlling VMI object, it should be limited to running
+stateless workloads.
+
+Removing the VMI object represents stopping the Virtual Machine.
+
+## Virtual Machine
+A [VirtualMachine](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachine)
+in the other hand, represents a VM that is not running or in a stopped state.
+
+The `VirtualMachine` holds the template from which to create the VirtualMachineInstance.
+
+By setting the 
+[VM.Spec](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachinespec)
+`running` property to `true`, the Virtual Machine is started (a corresponding
+VMI object is created).
+
+The status of a running VM should be retrieved from the associated VMI object,
+since the VM object does not provide runtime status.
+
+When the user deletes the associated VMI object, the KubeVirt controller just
+schedules another VMI - since the user has declared (via the `running`
+attribute) that the VM should be running.
+
+## Feature matrix per workload type
+
+|Feature name |Virtual Machine|Virtual Machine Instance|
+|-------------|---------------|------------------------|
+|LiveMigration|   supported   |      supported         |
+
+TODO: does it makes sense to have a table here ?... 
--- a/docs/virtual_machines/.pages
+++ b/docs/virtual_machines/.pages
@@ -1,4 +1,5 @@
 nav:
+  - vms_and_vmis.md
   - virtual_machine_instances.md
   - creating_vms.md
   - lifecycle.md


maiqueb:
/cc @aburdenthehand @cwilkers 

Would appreciate your feedback. 

... right now, there's not much meat on this. It's harder to write than I anticipated.

maiqueb:
@rmohr hey o/

Would you have some time to orient this new page ? 

Do you think a table with the supported feature set for each workload type is a good idea ? I mostly want to make it evident for users they miss out on something if they "choose" one of these types...

docs/virtual_machines/vms_and_vmis.md
```diff
@@ -0,0 +1,41 @@
...
+It is encapsulated in a pod.
+
+When used without a controlling VMI object, it should be limited to running
+stateless workloads.
+
+Removing the VMI object represents stopping the Virtual Machine.
+
+## Virtual Machine
+A [VirtualMachine](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachine)
+in the other hand, represents a VM that is not running or in a stopped state.
```
VirrageS, line 19: This is very confusing part.

Why does it represent VM that is not running? If I set it to running VirtualMachine no longer represents it?

docs/virtual_machines/vms_and_vmis.md
```diff
@@ -0,0 +1,41 @@
...
+
+The `VirtualMachine` holds the template from which to create the VirtualMachineInstance.
```
VirrageS, line 21: This seems like more introductory sentence than the one above.

docs/virtual_machines/vms_and_vmis.md
```diff
@@ -0,0 +1,41 @@
...
+
+By setting the 
+[VM.Spec](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachinespec)
+`running` property to `true`, the Virtual Machine is started (a corresponding
```
VirrageS, line 25:
```suggestion
`running` property to `true`, the virtual machine is started (a corresponding
```

docs/virtual_machines/vms_and_vmis.md
```diff
@@ -0,0 +1,41 @@
...
+VMI object is created).
+
+The status of a running VM should be retrieved from the associated VMI object,
+since the VM object does not provide runtime status.
+
+When the user deletes the associated VMI object, the KubeVirt controller just
+schedules another VMI - since the user has declared (via the `running`
+attribute) that the VM should be running.
+
```
VirrageS, line 34: It feels like some comparison is missing. Like when VMI should be used. Or when VM. What we gain from using just VMI and what when we use VM etc.

As a user I feel like this is a question I would like to be answered.

Simply which object should I use to create virtual machine.

docs/virtual_machines/vms_and_vmis.md
```diff
@@ -0,0 +1,41 @@
...
+A
+[VirtualMachineInstance](https://kubevirt.io/user-guide/virtual_machines/virtual_machine_instances/)
+represents a single **running** virtual machine.
+
+It is encapsulated in a pod.
+
+When used without a controlling VMI object, it should be limited to running
+stateless workloads.
+
+Removing the VMI object represents stopping the Virtual Machine.
```
VirrageS, line 15:
```suggestion
Removing the VMI object represents stopping the virtual machine.
```
Shouldnt we use VM in context of 'VirtualMachine' (Kubernetes object) and virtual machine as something that is started in a Pod?

VirrageS (changes requested):

docs/virtual_machines/vms_and_vmis.md
maiqueb: Good point. It's hard to define ... 

At the end of the day, I guess both are true... A running VM is - IMHO - represented by a VMI. Just look at its status object ...

Let me think on how to re-write this...

huge thanks for the feedback!

docs/virtual_machines/vms_and_vmis.md
maiqueb: Good point. I guess I can also tie this sentence neatly to the one above. 

i.e. a VM is a template for running VMs (VMIs) which in reality represents a stopped VM. 

All "live" VM things are available on the corresponding VMI object.

Makes sense ? 
@VirrageS

VirrageS: It makes sense. But I just wouldn't mention the "in reality represents a stopped vm". I feel like mentioning that VM is template for VMI which actually triggers creation of a Pod and in consequence a virtual machine is enough. 

I feel like there is no need to mention the state of the virtual machine. Unless just to point out that you can control the state of the VMI  (and effectively virtual machine) using VM spec.

docs/virtual_machines/vms_and_vmis.md
cwilkers: I like to explain / think of this in terms of a Kubernetes Deployment and its Pods. A Deployment can scale to many many Pods, but zero and one are also valid cases. In VMs, we have between zero and two at the most VMIs (migration case) and only one or zero running at a time.

Of course, it's somewhat more complicated with VM -> VMI -> virt-launcher Pod, but I think the analogy helps.

cwilkers (commented):
I've added some thoughts on how to frame things; also I wrote a [learning path](https://kubebyexample.com/learning-paths/kubevirt-fundamentals/overview-kubevirt) earlier this year on https://kubebyexample.com/

I'm especially happy with the [managing virtual machines](https://kubebyexample.com/learning-paths/kubevirt-fundamentals/managing-virtual-machines-kubevirt) page where I go through a VM yaml section by section.

Feel free to adapt any of that content (I'm inquiring about copyright now, so don't lift anything verbatim until we can verify it's copy-lefted)

PR #694, opened 2023-07-03
live-migration: no need of feature gates

victortoso:
Since v0.56, with the PR:
https://github.com/kubevirt/kubevirt/pull/8047

### DIFF
--- a/docs/operations/live_migration.md
+++ b/docs/operations/live_migration.md
@@ -6,7 +6,8 @@ continues to run and remain accessible.
 
 ## Enabling the live-migration support
 
-Live migration must be enabled in the feature gates to be supported. The
+Live migration is enabled by default in recent versions of KubeVirt. Versions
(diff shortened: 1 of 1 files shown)


aburdenthehand:
/lgtm
Although I do wonder, since v0.56 is now out of support, whether we just drop the whole para/section. 
Thanks @victortoso !

victortoso:
Hi @aburdenthehand anything else missing for this?

victortoso:
>  Although I do wonder, since v0.56 is now out of support, whether we just drop the whole para/section.

That's actually a good point. What lowest version we should keep in mind to update the user-guide?

aburdenthehand:
@victortoso as you know we talked a bit about this in the community meeting. Since this is version is only recently out of support, it makes sense to cover it the way that you have.
I've created an issue (#701) to update our contributor guide to include this kind of information.

vladikr:
Looks good to me. Thanks.
/approve

PR #699, opened 2023-07-12
Updating with v1.0 release notes with update_changelog.sh

aburdenthehand:
The update_changelog script strips the formatting from the tagged release notes.
After this is merged, a subsequent PR can return that formatting so that these are sectioned. For now, it is better that the release notes in some form are published in the user guide.

/cc @fabiand


phoracek (approved):
/lgtm
/approve

PR #704, opened 2023-08-01
Add documentation for persistent TPM and EFI

jean-edouard:


jobbler:
this document looks good to me. its easy to understand and looks pretty simple to do. I like it.
/lgtm

docs/virtual_machines/persistent_tpm_and_uefi_state.md
```diff
@@ -0,0 +1,85 @@
...
+  template:
+    spec:
+      domain:
+        devices:
+          tpm:
+            persistent: true
+```
+
+### Uses
+- The Microsoft Windows 11 installer requires the presence of a TPM device, even though it doesn't use this. Persistence is not required in this case however.
```
iholder101, line 51: what do you mean by "even though it doesn't use this"?

docs/virtual_machines/persistent_tpm_and_uefi_state.md
```diff
@@ -0,0 +1,85 @@
...
+### Limitations
+
+- As mentioned above, the backend storage PVC can only be created using a storage class that supports RWX FS. There is ongoing work to support block storage in future versions of KubeVirt.
+- Backend storage is currently incompatible with VM snapshot. It is planned to add snapshot support in the future.
+
+## TPM with persistent state
+
+Since KubeVirt v0.53.0, a TPM device can be added to a VM (with just `tpm: {}`). However, the data stored in it does not persist across reboots.  
+Support for persistence was added in v1.0.0 using a simple `persistent` boolean parameter that default to false, to preserve previous behavior.  
+Of course, backend storage must first be configured before adding a persistent TPM to a VM. See above.
```
iholder101, line 34: if it's not - will the VM creation be denied?

iholder101 (commented):
/approve
/hold

Thank you @jean-edouard!
Left two minor questions, which probably just help me personally to understand it a bit better :) feel entirely free to unhold if you don't think any further changes are necessary

docs/virtual_machines/persistent_tpm_and_uefi_state.md
jean-edouard: The installer forces you to have a TPM chip for no good reason. Neither it nor a freshly installed Windows will actually make use of the TPM (as long as you don't enable bitlocker).
So if you don't plan on using bitlocker, just add a non-persistent TPM to your VM and you'll be all set!
Do you think the docs need more info on this?

docs/virtual_machines/persistent_tpm_and_uefi_state.md
jean-edouard: That's a good question. IIRC it will stay pending until backend-storage is configured, but let me make sure!

jean-edouard: So, enabling backend-storage involves 2 things: adding the feature gate and populating the storage class.
Without the feature gate, the VM will fail to create.
Without the storage class, the VM will create, and starting it will create a pending VMI waiting for the storage class to be populated.
Do you think this is valuable info for the user docs?

docs/virtual_machines/persistent_tpm_and_uefi_state.md
iholder101: That's..... interesting? :D

Alright then, good to know!

dhiller (commented):
/approve

Thanks!

PR #706, opened 2023-08-15
cloud-init, ConfigDrive: Add networkData example

orelmisan:
When using the ConfigDrive datasource, the `networkData` part has to be in the [OpenStack Metadata Service Network](https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/metadata-service-network-info.html) format.

Add an example of such a configuration.

Tested with CentOS stream {8,9} and RHEL {8,9}.

### DIFF
--- a/docs/virtual_machines/startup_scripts.md
+++ b/docs/virtual_machines/startup_scripts.md
@@ -23,7 +23,8 @@ VMs.
 Cloud-init documentation can be found here: [Cloud-init
 Documentation](https://cloudinit.readthedocs.io/en/latest/).
 
-KubeVirt supports cloud-init's "NoCloud" and "ConfigDrive" datasources
+KubeVirt supports cloud-init's [NoCloud](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html) 
+and [ConfigDrive](https://cloudinit.readthedocs.io/en/latest/reference/datasources/configdrive.html) datasources
 which involve injecting startup scripts into a VM instance through the
 use of an ephemeral disk. VMs with the cloud-init package installed will
 detect the ephemeral disk and execute custom userdata scripts at boot.
--- a/docs/virtual_machines/startup_scripts.md
+++ b/docs/virtual_machines/startup_scripts.md
@@ -113,6 +113,32 @@ ConfigDrive data source:
         cloudInitConfigDrive:
           userData: "#cloud-config"
 
+When using the ConfigDrive datasource, the `networkData` part has to be in the
+[OpenStack Metadata Service Network](https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/metadata-service-network-info.html) format:
+```
+  spec:
+    domain:
+      interfaces: 
+        - name: secondary-net
+          bridge: {}
+          macAddress: '02:26:19:00:00:30'
+          model: virtio
+    networks: 
+        - multus:
+            networkName: my-ns/my-net
+          name: secondary-net
+    volumes:
+      - name: cloudinitvolume
+        cloudInitConfigDrive:
+          networkData: |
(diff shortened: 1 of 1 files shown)


docs/virtual_machines/startup_scripts.md
```diff
@@ -112,6 +113,17 @@ ConfigDrive data source:
         cloudInitConfigDrive:
           userData: "#cloud-config"
 
+When using the ConfigDrive datasource, the `networkData` part has to be in the
+[OpenStack Metadata Service Network](https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/metadata-service-network-info.html) format:
```
0xFelix, line 117: ... or in ENI (legacy) format, see https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-eni.html#network-config-eni.

docs/virtual_machines/startup_scripts.md
```diff
@@ -112,6 +113,17 @@ ConfigDrive data source:
...
+```
+       volumes:
+      - name: cloudinitvolume
+        cloudInitConfigDrive:
+          networkData: |
+            {"links":[{"id":"enp2s0","type":"phy","ethernet_mac_address":"02:26:19:00:00:30"}],"networks":[{"id":"NAD1","type":"ipv4","link":"enp2s0","ip_address":"10.184.0.244","netmask":"255.255.240.0","routes":[{"network":"0.0.0.0","netmask":"0.0.0.0","gateway":"23.253.157.1"}],"network_id":""}],"services":[]}
```
0xFelix, line 123: For a minimal example you may drop the empty `network_id` and the empty `services` list.

0xFelix (commented):
While the example works, can you add a note that matching of interfaces happens on the MAC address? Therefore you need to know MAC addresses in advance, before you can create a network configuration.

docs/virtual_machines/startup_scripts.md
orelmisan: The reason I kept `network_id` and `services` is that this format has a [JSON schema](https://docs.openstack.org/nova/latest/_downloads/9119ca7ac90aa2990e762c08baea3a36/network_data.json) to validate the data.

I think we should stick to the format, even though cloud-init is more [permissive](https://github.com/canonical/cloud-init/blob/7ffd0bcd8379ae6ece096e7014444978efa5f7c8/cloudinit/sources/helpers/openstack.py#L526).

docs/virtual_machines/startup_scripts.md
orelmisan: I don't think we should encourage using a deprecated format which we haven't tested.

orelmisan:
> While the example works, can you add a note that matching of interfaces happens on the MAC address? Therefore you need to know MAC addresses in advance, before you can create a network configuration.

Done.

dominikholler:
@lyarwood  can you please have a look?

docs/virtual_machines/startup_scripts.md
0xFelix: ACK, fair point.

docs/virtual_machines/startup_scripts.md
0xFelix: Yes, we don't want to encourage it, but it is supported. IMHO I don't see the OpenStack format being desirable either.

orelmisan: > Yes, we don't want to encourage it, but it is supported. 
Supported by cloud-init does not  necessarily mean supported by KubeVirt.

>  IMHO I don't see the OpenStack format being desirable either.
The reason for supporting this format is fact that credentials API works with `CloudInitConfigDrive` (at least for now).

orelmisan: > The documentation about this is quite unclear. Can you rework this part too? E.g. what is supported and under which circumstances? https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#network-config

I agree that this part should be reworked.
I can have a look at this part, but it will require more research.
I think it is out of the scope of this PR.

PR #708, opened 2023-08-21
Add note about writeback mode in disk cache

### LINKED ISSUE #627: Disk Device Cache mode update
Support for `writeback` caching was added in this PR: https://github.com/kubevirt/kubevirt/pull/6968

However the current docs do not reflect this: https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/#disk-device-cache

cwilkers:
Add to documentation that writeback is a valid cache mode for disks

Fixes #627

I did some research on the feature, but would appreciate sanity-check from some of the devs that I'm not misleading anybody here.

### DIFF
--- a/docs/virtual_machines/disks_and_volumes.md
+++ b/docs/virtual_machines/disks_and_volumes.md
@@ -1460,9 +1460,9 @@ KubeVirt supports `none`, `writeback`, and `writethrough` KVM/QEMU cache modes.
 -   `writeback` I/O from the guest is cached on the host and written through
     to the physical media when the guest OS issues a flush.
 
--   `writethrough` I/O from the guest is cached on the host but written
-    through to the physical medium on each write.
-
+-   `writethrough` I/O from the guest is cached on the host but must be written
+    through to the physical medium before the write operation completes.
+    
 > **Important:** `none` cache mode is set as default if the file system
 > supports direct I/O, otherwise, `writethrough` is used.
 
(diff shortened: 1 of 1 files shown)


cwilkers:
@ChandonPierre @adithyaakrishna @vladikr Can you PTAL? Thanks!

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -1451,14 +1451,17 @@ allocate 4 queues (to match the number of CPUs requested).
...
 -   `none` I/O from the guest is not cached on the host. Use this option
     for guests with large I/O requirements. This option is generally the
     best choice.
 
+-   `writeback` I/O from the guest is cached on the host and written through
+    to the physical media when the guest OS issues a flush.
+
 -   `writethrough` I/O from the guest is cached on the host but written
-    through to the physical medium.
+    through to the physical medium on each write.
```
vladikr, line 1464: I'm not 100% sure how to phrase it best. Perhaps we should say that the data is stored in the page cache on the host and writes are reported as completed to the guest only when the data has been committed to the storage device.

@alicefr what do you think?

alicefr: @vladikr yes, it sounds a more correct formulation to me. Here, we are missing that the writes are completed when they are done on the storage device.

cwilkers: I updated the wording, is this more correct?

alicefr: Yes, sounds good to me

vladikr:
Looks good, thanks!
/lgtm

cwilkers:
Thanks!
Since I've got two lgtms, I'll approve my own PR:
/approve

PR #710, opened 2023-08-25
Pointing to more appropriate contributing page

### LINKED ISSUE #659: Redirect users to Contributing Guide Directly
**Description:**

Instead of directing users from https://github.com/kubevirt/user-guide/blob/main/CONTRIBUTING.md to https://github.com/kubevirt/community/blob/main/contributors/contributing.md and from there to https://kubevirt.io/user-guide/appendix/contributing/ - ( which is currently giving a 404 Error. I have fixed the link and created a PR for the same here, https://github.com/kubevirt/community/pull/214 ), how about we do it directly from the user-guide to https://kubevirt.io/user-guide/contributing/

Another option, we can redirect them directly to https://github.com/kubevirt/kubevirt/blob/main/CONTRIBUTING.md

So the users instead of traversing through the maze, will be taken directly from user-guide to https://kubevirt.io/user-guide/contributing/ 

Fixed by  #710

aburdenthehand:
Previously pointed another page that in turn links to the user guide contributing page. 

Fixes #659

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
 # Contributing guidelines
 
-Please follow the same [contributing guide](https://github.com/kubevirt/community/blob/main/contributors/contributing.md) as you would follow for any other pull request made to Kubevirt.
+Please see the [contributing guide](https://kubevirt.io/user-guide/contributing) for information on how and where to contribute to the project.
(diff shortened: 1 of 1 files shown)


phoracek (approved):
/lgtm
/approve

PR #711, opened 2023-09-01
Fix indentation of features stanza in BIOS/UEFI

jamesandariese:
`features` is currently nexted under devices in the docs for BIOS/UEFI.

It should be alongside devices rather than under it.


cwilkers:
This matches what I have on my system, thanks for the PR, good catch!
/approve
/lgtm

PR #713, opened 2023-09-05
Document memory and CPU requests and limits

jean-edouard:


docs/virtual_machines/resources_requests_and_limits.md
```diff
@@ -0,0 +1,59 @@
...
+metadata:
+  labels:
+    autoCpuLimit: "true"
+    kubernetes.io/metadata.name: default
+  name: default
+```
+
+## Memory
+### Memory requests on the container
+- VM(I)s must specify a desired amount of memory, in either spec.domain.memory.guest or spec.domain.resources.requests.memory (ignoring hugepages, see the [dedicated page](../operations/hugepages.md)). If both are set, the memory requests take precedence. The container will request that amount + a calculated overhead.
```
dhiller, line 51: I'd suggest replacing the `+ ....` with `, adding a calculated overhead` or `, including an added calculated overhead`

docs/virtual_machines/resources_requests_and_limits.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Memory limits on the container
+- By default, no memory limit is set on the container
+- Manually setting a memory limit on the VM(I) will set the same value on the container
+
+#### Warnings
+- Memory limits have to be more than memory requests + overhead
```
dhiller, line 58: Maybe we should describe what happens in that case.

dhiller (commented):
/approve

Thank you @jean-edouard !

Two nits inline.

jean-edouard:
@dhiller thank you for the review! I tried to address your comments, English is hard... PTAL!

vladikr:
Looks great!

dhiller (approved):
Thank you, this version is even better!

PR #721, opened 2023-10-02
instancetype: Add view ClusterRole docs

lyarwood:


docs/virtual_machines/instancetypes.md
```diff
@@ -469,6 +469,32 @@ spec:
 EOF
 ```
 
+## `instancetype.kubevirt.io:view` `ClusterRole`
+
+From version `v1.1.0` a stand alone `instancetype.kubevirt.io:view` [`ClusterRole`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) is provided to allow users read only to access `VirtualMachineCluster{Instancetype,Preference}` resources.
+
+[Bug #10437](https://github.com/kubevirt/kubevirt/issues/10437) sets out the justification for this `ClusterRole` and why the [default aggregated `view` `ClusterRole`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) isn't enough when bound to a user through a [`RoleBinding`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding). Binding a user to the default `view` `ClusterRole` with a `ClusterRoleBinding` would obviously provide a user with read only access to all resources aggregated to the role across all namespaces and is thus not a suitable option.
```
aburdenthehand, line 476: I would much prefer we include the information here instead of linking to a bug. Rather than justification, I think a user would be more interested in when and why they should use this role.
Maybe something along the lines of "The `instance.type.kubevirt.io:view` cluster role allows users to view all instance type and preference resources in the cluster but does not otherwise provide access to other cluster-wide resources, making it useful for VM users rather than administrators." 
?

aburdenthehand (commented):
Thanks @lyarwood! 
A comment and a couple of suggestions from me.

docs/virtual_machines/instancetypes.md
lyarwood: ACK I've rewritten this without a link to the bug.

0xFelix (commented):
Looks good, just two notes and a question.

Can you mention this role could be bound automatically to users depending on the authentication provider in use, similar to the `view` ClusterRole?

lyarwood:
/close

This is no longer required after https://github.com/kubevirt/kubevirt/pull/11025 that adds the role automatically for all users.

PR #718, opened 2023-09-19
Add USB pass through documentation

victortoso:
Hi,

This is related to the feature that is not yet merged: https://github.com/kubevirt/kubevirt/pull/10015

Cheers,

PR #723, opened 2023-10-10
Enabling footnotes feature

aburdenthehand:
Adding the [footnotes features](https://squidfunk.github.io/mkdocs-material/reference/footnotes/) for material

As seen in #718 footnotes can be pretty useful in a place where an admonition might be obtrusive.


dhiller (approved):
/approve

PR #735, opened 2023-11-15
Document how to use the error policy

alicefr:
Document the option introduced by https://github.com/kubevirt/kubevirt/pull/10058 to configure the error policy for disks and luns


cwilkers:
Hi @alicefr thanks for the PR; it looks in order to me, do you want any other reviewers? 
/lgtm

alicefr:
> Hi @alicefr thanks for the PR; it looks in order to me, do you want any other reviewers? /lgtm

I think the change is relatively small and for a single field that it is enough, wdyt?

cwilkers:
> I think the change is relatively small and for a single field that is enough, wdyt?

Okay, I'll put in the approval
/approve

PR #736, opened 2023-11-21
IsolateEmulatorThread with full-pcpus-only

orelmisan:
Following kubevirt/kubevirt#10593 it is possible to run VMs with dedicated CPU placement and emulator thread isolation on compute nodes with SMT [1] enabled and `full-pcpus-only` CPUManager policy [2].

This configuration is used for high-performance VMs.

Add an explanation on how to use it.

[1] https://en.wikipedia.org/wiki/Simultaneous_multithreading
[2] https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy-options

### DIFF
--- a/docs/virtual_machines/dedicated_cpu_resources.md
+++ b/docs/virtual_machines/dedicated_cpu_resources.md
@@ -114,6 +114,26 @@ Example:
             cpu: 2
             memory: 2Gi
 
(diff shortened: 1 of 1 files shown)


RamLavi (approved):
/lgtm

PR #740, opened 2023-11-25
cloneAPI: add how to use template label and annotation filters

matthewei:
As we have add template label and annotation filters and we want to add how to use template label and annotation filters.

### DIFF
--- a/docs/operations/clone_api.md
+++ b/docs/operations/clone_api.md
@@ -108,7 +108,7 @@ Setting label / annotation filters is **optional**. If unset, all labels / annot
 
 #### Template Label & Template Annotation filters
 
-Some network cni such as kube-ovn or ovn-kubernetes will inject network information into the annotations of the VM. It would raise an issue When we want to clone a vm from the target vm. The clone vm would use the same network. So you can use template labels and annotation filters to avoid this problems.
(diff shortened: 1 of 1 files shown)


docs/operations/clone_api.md
```diff
@@ -98,6 +106,10 @@ used:
 
 Setting label / annotation filters is **optional**. If unset, all labels / annotations will be copied as a default.
 
+#### Template Label & Template Annotation filters
+
+Some network cni such as kube-ovn or ovn-kubernetes will inject network information into the annotations of the VM. It would raise an issue When we want to clone a vm from the target vm. The clone vm would use the same network. So you can use template labels and annotation filters to avoid this problems.
```
dhiller, line 111:
```suggestion
Some network CNIs such as Kube-OVN or OVN-Kubernetes inject network information into the annotations of a VM. When cloning a VM from a target VM the cloned VM will use the same network. To avoid this you can use template labels and annotation filters.
```

dhiller (changes requested):
Thanks for updating the user guide!

As a nit I have reworded a bit, hope it makes sense.

docs/operations/clone_api.md
matthewei: yes, Thanks a lot! I will follow your suggestion. Many thanks!

matthewei:
> Thanks for updating the user guide!
> 
> As a nit I have reworded a bit, hope it makes sense.

done

PR #744, opened 2023-12-12
Revert "IsolateEmulatorThread with full-pcpus-only (#736)"

orelmisan:
Revert changes made by PR #736.
PR https://github.com/kubevirt/community/pull/247 now proposes a different approach.


RamLavi (approved):
/lgtm

after we introduce the global knob - we can update it here on a separate PR

PR #750, opened 2024-01-03
Clarify raw size mismatch in export documentation

alromeros:
CDI resizes raw imports to match the size requested for the PVC. This PR adds some clarifications to the export documentation to specify why sometimes raw image size may be larger than expected.

Fixes: https://issues.redhat.com/browse/CNV-36303

### DIFF
--- a/docs/operations/export_api.md
+++ b/docs/operations/export_api.md
@@ -252,7 +252,7 @@ status:
 #### Format types
 There are 4 format types that are possible:
 
-* Raw. The unaltered raw KubeVirt disk image.
+* Raw. The unaltered raw KubeVirt disk image. Raw images may be larger than expected if they were previously resized to match the PVC requested size.
 * Gzip. The raw KubeVirt disk image but gzipped to help with transferring efficiency.
 * Dir. A directory listing, allowing you to find the files contained in the PVC.
 * Tar.gz The contents of the PVC tarred and gzipped in a single file.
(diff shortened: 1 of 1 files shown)


docs/operations/export_api.md
```diff
@@ -252,7 +252,7 @@ status:
 #### Format types
 There are 4 format types that are possible:
 
-* Raw. The unaltered raw KubeVirt disk image.
+* Raw. The unaltered raw KubeVirt disk image. Raw images may be larger than expected if they were previously resized to match the PVC requested size.
```
awels, line 255: I would note something about the virtual size vs the actual size. A standard ls will report the virtual size of a sparse image. This is why it might look larger, but won't really be that much larger when you look at the actual size

alromeros: That's right... However, since in practice we always show the actual size during the download, I guess that could lead to confussion if then we want to upload the image to a PVC with a smaller virtual size as described in https://issues.redhat.com/browse/CNV-36303 . I agree with you though, I'm not sure if this is the right documentation to specify this. 

What about something like this:

```md
* Raw. The unaltered raw KubeVirt disk image. Raw images' virtual size may be larger than expected if they were previously resized to match the PVC requested size.
```

awels: The reason I don't like these type of messages is that even if we uploaded it to a PVC of the exact size the virtual size was, and thus nothing changed. ls will still show the virtual size which is not the actual size. This is all perfectly normal behavior, so pointing it out seems pointless.

alromeros:
We decided this is not the right place to specify the possible differences between virtual and actual sizes. Closing.

PR #751, opened 2024-01-05
User guide for hook sidecar

### LINKED ISSUE #726: Hook sidecar documentation on user guide
https://github.com/kubevirt/kubevirt/pull/10479 adds support to run scripts as a part of hook sidecar. The PR adds docs which are available on kubevirt repo, but nothing about sidecars exists on user guide (at least I couldn't find anything around using it). If it makes sense, I'd like to add documentation for this on the user-guide since, I think, it's something that should be in the user facing docs.

If this sounds reasonable, I'd need some help figuring out which section of user guide this would fit in - a separate section of its own under "Operations" tab, or under something existing?

dharmit:
Fixes #726

### DIFF
--- a/docs/operations/hook-sidecar.md
+++ b/docs/operations/hook-sidecar.md
@@ -1,0 +1,168 @@
+# Hook Sidecar Container
+
+## Introduction
+
+In KubeVirt, a Hook Sidecar container is a sidecar container (a secondary container that runs along with the main
+application container within the same Pod) used to apply customizations before the Virtual Machine is initialized. This
+ability is provided since 100% of the [libvirt domain XML](https://libvirt.org/formatdomain.html) elements and
+attributes are not configurable/customizable using the VMI specification.
+
+The sidecar containers communicate with the main container over a socket with a gRPC protocol. There are two main
+sidecar hooks:
+
+1. `onDefineDomain`: This hook helps to customize libvirt's XML and return the new XML over gRPC for the VM creation.
+2. `preCloudInitIso`: This hook helps to customize the cloud-init configuration. It operates on and returns JSON
+   formatted cloud-init data.
+
+## Enabling Sidecar feature gate
+
+To be able to use the sidecar hooks, the Sidecar feature gate must be enabled on the cluster. If you're using a
+development cluster backed by KubeVirt CI, you could enable it using the following steps:
+```shell
+# export FEATURE_GATES=<feature-gate-1>,<feature-gate-2>
+# e.g. to enable Sidecar and HotplugNICs feature gates run below
+$ export FEATURE_GATES=Sidecar,HotplugNICs
+$ make cluster-sync
+```
+
+For non-development clusters, it can be enabled by following the steps mentioned in [Activating feature
+gates](../activating_feature_gates).
+
+## Sidecar-shim container image
+
+To reduce the amount of boilerplate that developers need to do in order to run VM with custom modifications, we provide
+the [sidecar-shim-image](https://quay.io/repository/kubevirt/sidecar-shim) that takes care of implementing the
+communication with the main container.
+
+The image contains the `sidecar-shim` binary built using
+[`sidecar_shim.go`](https://github.com/kubevirt/kubevirt/blob/main/cmd/sidecars/sidecar_shim.go) which should be kept
+as the entrypoint of the container. This binary will search in `$PATH` for binaries named after the hook names (e.g
+`onDefineDomain` and `preCloudInitIso`) and run them. User must provide the necessary arguments as command line options
+(flags).
+
+In the case of `onDefineDomain`, the arguments will be the VMI information as JSON string, (e.g `--vmi vmiJSON`) and
+the current domain XML (e.g `--domain domainXML`). As standard output it expects the modified domain XML.
+
+In the case of `preCloudInitIso`, the arguments will be the VMI information as JSON string, (e.g `--vmi vmiJSON`) and
+the CloudInitData (e.g `--cloud-init cloudInitJSON`). As standard output it expects the modified CloudInitData (as
+JSON).
+
+Besides a binary, one could also execute shell or python scripts by making them available at the expected location.
+
+## Go, Python, Shell - pick any one
+
+Although a binary doesn't strictly need to be generated from Go code, and a script doesn't strictly need to be one
+among Shell or Python, for the purpose of this guide, we will use those as examples.
+
+### Go binary
+
+Example Go code modifiying the [SMBIOS system
+information](https://libvirt.org/formatdomain.html#smbios-system-information) can be found in the [KubeVirt
+repo](https://github.com/kubevirt/kubevirt/tree/main/cmd/sidecars/smbios). Binary generated from this code, when
+available under `/usr/bin/ondefinedomain` in the sidecar-shim-image, is run right before VMI creation and the baseboard
+manufacturer value is modified to reflect what's provided in the `smbios.vm.kubevirt.io/baseBoardManufacturer`
+annotation in [VMI spec](https://github.com/kubevirt/kubevirt/blob/main/examples/vmi-with-sidecar-hook.yaml).
+
+### Shell or Python script
+
+If you pefer writing a shell or python script instead of a Go program, create a Kubernetes ConfigMap and use
+annotations to make sure the script is run before the VMI creation. The flow would be as below:
+
+1. Create a ConfigMap containing the shell or python script you want to run
+1. Create a VMI containing the annotation `hooks.kubevirt.io/hookSidecars` and mention the ConfigMap information in it.
+
+#### ConfigMap with shell script
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-config-map
+data:
+  my_script.sh: |
+    #!/bin/sh
+    tempFile=`mktemp --dry-run`
+    echo $4 > $tempFile
+    sed -i "s|<baseBoard></baseBoard>|<baseBoard><entry name='manufacturer'>Radical Edward</entry></baseBoard>|" $tempFile
+    cat $tempFile
+```
+
+#### ConfigMap with python script
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-config-map
+data:
+  my_script.sh: |
+    #!/usr/bin/env python
+
+    import xml.etree.ElementTree as ET
+    import sys
+
+    def main(s):
+        # write to a temporary file
+        f = open("/tmp/orig.xml", "w")
+        f.write(s)
+        f.close()
+
+        # parse xml from file
+        xml = ET.parse("/tmp/orig.xml")
+        # get the root element
+        root = xml.getroot()
+        # find the baseBoard element
+        baseBoard = root.find("sysinfo").find("baseBoard")
+
+        # prepare new element to be inserted into the xml definition
+        element = ET.Element("entry", {"name": "manufacturer"})
+        element.text = "Radical Edward"
+        # insert the element
+        baseBoard.insert(0, element)
+
+        # write to a new file
+        xml.write("/tmp/new.xml")
+        # print file contents to stdout
+        f = open("/tmp/new.xml")
+        print(f.read())
+        f.close()
+
+    if __name__ == "__main__":
+        main(sys.argv[4])
+```
+
+After creating one of the above ConfigMap, create the VMI using the manifest in [this
+example](https://github.com/kubevirt/kubevirt/blob/main/examples/vmi-with-sidecar-hook-configmap.yaml). Of importance
+here is the ConfigMap information stored in the annotations:
+
+```yaml
+annotations:
+  hooks.kubevirt.io/hookSidecars: >
+    [
+        {
+            "args": ["--version", "v1alpha2"],
+            "image": "registry:5000/kubevirt/sidecar-shim:devel",
+            "configMap": {"name": "my-config-map", "key": "my_script.sh", "hookPath": "/usr/bin/onDefineDomain"}
+        }
+    ]
+```
+
+
+The `name` field indicates the name of the ConfigMap on the cluster which contains the script you want to execute. The
+`key` field indicates the key in the ConfigMap which contains the script to be executed. Finally, `hookPath` indicates
+the path where you want the script to be mounted. It could be either of `/usr/bin/onDefineDomain` or
+`/usr/bin/preCloudInitIso` depending upon the hook you want to execute.
+
+### Verify everything works
+
+Whether you used the Go binary or a Shell/Python script from above examples, you would be able to see the newly created
+VMI have the modified baseboard manufacturer information. After creating the VMI, verify that it is in the `Running`
+state, and connect to its console and see if the desired changes to baseboard manufacturer get reflected:
+
+```shell
+# Once the VM is ready, connect to its display and login using name and password "fedora"
+cluster/virtctl.sh vnc vmi-with-sidecar-hook-configmap
+
+# Check whether the base board manufacturer value was successfully overwritten
+sudo dmidecode -s baseboard-manufacturer
+```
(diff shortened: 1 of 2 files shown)


docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Introduction
+
+In KubeVirt, a Hook Sidecar container is a sidecar container (a secondary container that runs along with the main
+application container within the same Pod) used to apply changes before the Virtual Machine is initialized. This
+ability is provided since 100% of the [libvirt domain XML](https://libvirt.org/formatdomain.html) elements and
+attributes are not configurable/customizable using the VMI specification.
+
+The sidecar containers communicate with the main container over a socket with a gRPC protocol. It is meant to do the
+changes over libvirt's XML and return the new XML over gRPC for the VM creation.
```
victortoso, line 11: > It is meant to do the changes over libvirt's XML and return the new XML over gRPC for the VM creation.

This refers only to the `onDefineDomain` hook but we also have the `preCloudInitIso` which does not requite libvirt's domain changes.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,145 @@
+# Hook Sidecar Container
+
+## Introduction
+
+In KubeVirt, a Hook Sidecar container is a sidecar container (a secondary container that runs along with the main
+application container within the same Pod) used to apply changes before the Virtual Machine is initialized. This
```
victortoso, line 6: >  apply changes before the Virtual Machine is initialized

I'd use `apply customizations` instead

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,145 @@
+# Hook Sidecar Container
+
+## Introduction
```
victortoso, line 3: In the introduction, you can point out what `preCloudInitIso` is used for too

victortoso (commented):
Thanks @dharmit! I might be a little biased on what is needed in the docs for someone that is just trying the hook sidecar. Still, I think it is a good start.

oreillymj:
I know this sounds dumb, but the documentation would benefit from have the instructions on how to enable the sidecar feature gate, especially since it's case sensitive. The doc doesn't say it's behind a feature gate so users may assume that it'll work by default, which is obviously not the case.
Is my spelling in the yaml below is correct?

```
cat << END > enable-sidecar-feature-gate.yaml
---
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  name: kubevirt
  namespace: kubevirt
spec:
  configuration:
    developerConfiguration: 
      featureGates:
        - Sidecar
END

kubectl apply -f enable-sidecar-feature-gate.yaml
```

docs/operations/hook-sidecar.md
dharmit: Would it be right to describe `preCloudInitIso` hook as one that modifies the cloud-init data before spinning up the VMI?

docs/operations/hook-sidecar.md
dharmit: Same as https://github.com/kubevirt/user-guide/pull/751#discussion_r1446029755. I'm myself having a hard time understanding what _exactly_ it does.

dharmit:
> I know this sounds dumb, but the documentation would benefit from have the instructions on how to enable the sidecar feature gate, especially since it's case sensitive. The doc doesn't say it's behind a feature gate so users may assume that it'll work by default, which is obviously not the case.

I don't think it's dumb. But we can put a link to [feature gate section](https://kubevirt.io/user-guide/operations/activating_feature_gates/#how-to-activate-a-feature-gate) and say that the feature gate `Sidecar` needs to be enabled to use this.

docs/operations/hook-sidecar.md
victortoso: Right, but I think you can't escape mentioning/reference to [cloud-init](https://github.com/kubevirt/kubevirt/tree/main/pkg/cloud-init) (... which is missing a README.md file!)

docs/operations/hook-sidecar.md
victortoso: Exactly, that's why we need documentation :)
I'm sorry, i can't help you too much here. I didn't look at it too closely, only what the [cloudinit example](https://github.com/kubevirt/kubevirt/blob/main/cmd/sidecars/cloudinit/cloudinit.go) sidecar does plus testing it.

dharmit:
@victortoso WDYT about the docs I have added for precloudinitiso?

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,162 @@
...
+In KubeVirt, a Hook Sidecar container is a sidecar container (a secondary container that runs along with the main
+application container within the same Pod) used to apply customizations before the Virtual Machine is initialized. This
+ability is provided since 100% of the [libvirt domain XML](https://libvirt.org/formatdomain.html) elements and
+attributes are not configurable/customizable using the VMI specification.
+
+The sidecar containers communicate with the main container over a socket with a gRPC protocol. There are two main
+sidecar hooks:
+1. `onDefineDomain`: This hook helps to customize libvirt's XML and return the new XML over gRPC for the VM creation.
+2. `preCloudInitIso`: This hook helps to customize the cloud-init configuration. It operates on and returns JSON
+   formatted cloud-init data.
```
victortoso, line 14: This did not break down as a list in the end result page. I guess it misses an empty line between the list and previous paragraph.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,162 @@
...
+
+After creating one of the above ConfigMap, create the VMI using the manifest in [this
+example](https://github.com/kubevirt/kubevirt/blob/main/examples/vmi-with-sidecar-hook-configmap.yaml). Of importance
+here is the ConfigMap information stored in the annotations:
+
+```yaml
+annotations:
+  hooks.kubevirt.io/hookSidecars: '[{"args": ["--version", "v1alpha2"],
+    "image": "registry:5000/kubevirt/sidecar-shim:devel",
+    "configMap": {"name": "my-config-map", "key": "my_script.sh", "hookPath": "/usr/bin/onDefineDomain"}}]'
```
victortoso, line 141: The other day, I found [this example](https://github.com/kubernetes/kubernetes/pull/19758/commits/c8c82c1d8f8e143f9b571a731d7ff11d5101de20) as to how break down json annotations in a better format.

I just tested this with `examples/vmi-with-sidecar-hook.yaml`, the diff is:
```diff
-    hooks.kubevirt.io/hookSidecars: '[{"args": ["--version", "v1alpha2"], "image":
-      "registry:5000/kubevirt/example-hook-sidecar:devel"}]'
+    hooks.kubevirt.io/hookSidecars: >
+      [
+        {
+          "args": ["--version", "v1alpha2"],
+          "image": "registry:5000/kubevirt/example-hook-sidecar:devel"
+        }
+      ]
```

what do you think, about using this format instead?

victortoso (commented):
> @victortoso WDYT about the docs I have added for precloudinitiso?

IMHO, looks fine. I've added two nits about the format

docs/operations/hook-sidecar.md
dharmit: Thanks for the catch. Fixed.

docs/operations/hook-sidecar.md
dharmit: Wow! Amended.

victortoso:
/lgtm
Thanks!

dharmit:
@aburdenthehand @cwilkers can one of you PTAL? I don't know who else to ping on this.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
...
+#### ConfigMap with python script
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-config-map
+data:
+  my_script.sh: |
+    #!/usr/bin/env python
```
stefanha, line 99: The sidecar-shim image has `python3` but not `python`. @victortoso and I noticed this when debugging together. Just a reminder to update this.

cwilkers (changes requested):
Just one sentence reorder, otherwise it looks good to me.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+The `name` field indicates the name of the ConfigMap on the cluster which contains the script you want to execute. The
+`key` field indicates the key in the ConfigMap which contains the script to be executed. Finally, `hookPath` indicates
+the path where you want the script to be mounted. It could be either of `/usr/bin/onDefineDomain` or
+`/usr/bin/preCloudInitIso` depending upon the hook you want to execute.
+
+### Verify everything works
+
+Whether you used the Go binary or a Shell/Python script from above examples, you would be able to see the newly created
```
aburdenthehand, line 158:
```suggestion
Whether you used the Go binary or a Shell/Python script from the above examples, you would be able to see the newly created
```

aburdenthehand (commented):
@dharmit Great stuff.
I have a few suggestions and questions for you to consider.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+For non-development clusters, it can be enabled by following the steps mentioned in [Activating feature
+gates](../activating_feature_gates).
+
+## Sidecar-shim container image
+
+To reduce the amount of boilerplate that developers need to do in order to run VM with custom modifications, we provide
+the [sidecar-shim-image](https://quay.io/repository/kubevirt/sidecar-shim) that takes care of implementing the
+communication with the main container.
```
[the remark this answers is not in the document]
dharmit: IMO, the reword that you have suggested doesn't accurately reflect what sidecar-shim image is intended for.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
...
+
+The image contains the `sidecar-shim` binary built using
+[`sidecar_shim.go`](https://github.com/kubevirt/kubevirt/blob/main/cmd/sidecars/sidecar_shim.go) which should be kept
+as the entrypoint of the container. This binary will search in `$PATH` for binaries named after the hook names (e.g
+`onDefineDomain` and `preCloudInitIso`) and run them. User must provide the necessary arguments as command line options
+(flags).
+
+In the case of `onDefineDomain`, the arguments will be the VMI information as JSON string, (e.g `--vmi vmiJSON`) and
+the current domain XML (e.g `--domain domainXML`). As standard output it expects the modified domain XML.
```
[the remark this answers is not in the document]
dharmit: Yes, it sounds the same to me. Do you suggest changing it what you asked? Why? They already sound the same thing to me.

NOTE: I'm not a native English speaker. :)

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
...
+
+In the case of `preCloudInitIso`, the arguments will be the VMI information as JSON string, (e.g `--vmi vmiJSON`) and
+the CloudInitData (e.g `--cloud-init cloudInitJSON`). As standard output it expects the modified CloudInitData (as
+JSON).
+
+Besides a binary, one could also execute shell or python scripts by making them available at the expected location.
```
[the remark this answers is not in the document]
dharmit: > What is the expected location, and does it make sense to say that location instead?

It does make sense to do so. For the suggestion, I have the same question as above (https://github.com/kubevirt/user-guide/pull/751#discussion_r1503978407).

dharmit:
@aburdenthehand thanks for taking the time to review and for your suggestions. I'm not a native English speaker, but I like writing docs. So the questions I have asked are out of genuine curiosity to learn more from a native person. :)

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
+# Hook Sidecar Container
+
+## Introduction
+
+In KubeVirt, a Hook Sidecar container is a sidecar container (a secondary container that runs along with the main
+application container within the same Pod) used to apply customizations before the Virtual Machine is initialized. This
+ability is provided since 100% of the [libvirt domain XML](https://libvirt.org/formatdomain.html) elements and
```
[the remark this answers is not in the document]
dharmit: Thanks, @cwilkers. Done.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
...
+attributes are not configurable/customizable using the VMI specification.
```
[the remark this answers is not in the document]
dharmit: Amended with what @cwilkers suggested. PTAL if it makes sense now.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,168 @@
...
+sidecar hooks:
+
+1. `onDefineDomain`: This hook helps to customize libvirt's XML and return the new XML over gRPC for the VM creation.
+2. `preCloudInitIso`: This hook helps to customize the cloud-init configuration. It operates on and returns JSON
+   formatted cloud-init data.
+
+## Enabling Sidecar feature gate
+
+To be able to use the sidecar hooks, the Sidecar feature gate must be enabled on the cluster. If you're using a
+development cluster backed by KubeVirt CI, you could enable it using the following steps:
```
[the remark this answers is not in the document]
dharmit: Makes sense. I have flipped the order and shortened it to a link to developer doc instead of the details.

docs/operations/hook-sidecar.md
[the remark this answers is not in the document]
aburdenthehand: From how I read it, "As a standard output" is an adverbial phrase. It is the broader category in which "XML" is one of several options. It adds additional context to the sentence, and, as such, having it start the sentence makes it a bit awkward to read because the reader has to remember the context until it reaches the part of the sentence to which it applies. 

I'd say the best way to write this is: "It expects modified domain XML as standard output." (The subject > verb > object > context) But what the "It" is is not entirely clear. Is it the binary?

docs/operations/hook-sidecar.md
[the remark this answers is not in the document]
dharmit: Yes, that's correct. And to that extent your suggestion seems appropriate. However, I think I'm stuck between whether the reader knows to know the 'boilerplate' thing or not. It's debatable, I guess. But we can do without it.

I'll modify it based on your suggestion.

docs/operations/hook-sidecar.md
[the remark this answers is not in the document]
dharmit: Thanks for the explanation.

> I'd say the best way to write this is: "It expects modified domain XML as standard output." (The subject > verb > object > context) But what the "It" is is not entirely clear. Is it the binary?

The "it" is the 'onDefineDomain' hook. We are starting the paragraph with `In the case of onDefineDomain`. So, to me, it seems that the "it" part is clear. Let me know if you think being explicit would be better.

aburdenthehand:
/lgtm
Thanks @dharmit !

dharmit:
@aburdenthehand @cwilkers can one of give an ok-to-test on this PR?

cwilkers:
> @aburdenthehand @cwilkers can one of give an ok-to-test on this PR?

/ok-to-test

It recently came to my attention that ok-to-test really has no effect on the docs repositories since our CI process here basically just creates previews and runs within Netlify's service.

docs/operations/hook-sidecar.md
```diff
@@ -0,0 +1,162 @@
...
+The sidecar containers communicate with the main container over a socket with a gRPC protocol. There are two main
+sidecar hooks:
+
+1. `onDefineDomain`: This hook helps to customize libvirt's XML and return the new XML over gRPC for the VM creation.
+2. `preCloudInitIso`: This hook helps to customize the cloud-init configuration. It operates on and returns JSON
+   formatted cloud-init data.
+
+## Enabling Sidecar feature gate
+
+Sidecar feature gate can be enabled by following the steps mentioned in [Activating feature gates](..
```
cwilkers, line 19: It would be helpful to add the exact capitalization/spelling for the feature gate in backtick quotes so the reader knows precisely what they are enabling, even if that is `Sidecar`.

dharmit: Sure. I have modified in the section title as well. PTAL.

PR #756, opened 2024-01-22
Add Memory Hotplug Documentation

acardace:
Add user-guide page about memory hotplug.

### DIFF
--- a/docs/operations/memory_hotplug.md
+++ b/docs/operations/memory_hotplug.md
@@ -1,0 +1,136 @@
+# Memory Hotplug
+
+Memory hotplug was introduced in KubeVirt version 1.1, enabling the dynamic resizing of the amount of memory available to a running VM.
+
+## Limitations
+* Memory hotplug is currently only supported on the x86_64 architecture.
+* Current hotplug implementation involves live-migration of the VM workload.
+
+# Configuration
+
+### Enable feature-gate
+
+To use memory hotplug we need to add the `VMLiveUpdateFeatures` feature gate in the KubeVirt CR:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - VMLiveUpdateFeatures
+```
+
+### Configure the Workload Update Strategy
+
+Configure `LiveMigrate` as `workloadUpdateStrategy` in the KubeVirt CR, since the current implementation of the hotplug process requires the VM to live-migrate.
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+spec:
+  workloadUpdateStrategy:
+    workloadUpdateMethods:
+    - LiveMigrate
+```
+
+### Enable in VM Specification
+
+Finally, enable memory hotplug in the VM specification:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+spec:
+  liveUpdateFeatures:
+    memory:
+      maxGuest: 2Gi
+```
+Specifying the memory.maxGuest value is optional. If left unset, it defaults to the value defined in maxHotplugRatio inside the KubeVirt CR (refer to [KubeVirt API Reference](https://kubevirt.io/api-reference/v1.1.0/definitions.html#_v1_liveupdateconfiguration)). For example, if a VM is configured with 512Mi of guest memory and maxGuest is not defined, and maxHotplugRatio is 2, then maxGuest will equal 1Gi.
+
+**NOTE:** If memory hotplug is enabled/disabled on an already running VM a reboot is necessary for the changes to take effect.
+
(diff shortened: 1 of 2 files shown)


dhiller:
/uncc @mazzystr 
/cc @aburdenthehand 

Andrew, do you mind taking a look also! Thank you!

dhiller (commented):
/approve

fossedihelm (approved):
Thanks @acardace! Looks good! Just a small nit with question.
Feel free to unhold if you think that it is not worth it or it is wrong
/hold

docs/operations/memory_hotplug.md
```diff
@@ -0,0 +1,136 @@
+# Memory Hotplug
+
+Memory hotplug was introduced in KubeVirt version 1.1, enabling the dynamic resizing of the amount of memory available to a running VM.
```
aburdenthehand, line 3: It would be good to separate the feature description from the version it was introduced into separate sentences, possibly even paras. That way we won't have to rewrite it when the version is no longer relevant.
And this might possibly be a nit as maybe there's enough information but is there anything else we could add to the feature description for someone that might be completely new to memory hotplugging?

docs/operations/memory_hotplug.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Configuration
+
+### Enable feature-gate
+
+To use memory hotplug we need to add the `VMLiveUpdateFeatures` feature gate in the Kubevirt CR:
```
aburdenthehand, line 9:
```suggestion
To use memory hotplug we need to add the `VMLiveUpdateFeatures` feature gate in the KubeVirt CR:
```

docs/operations/memory_hotplug.md
```diff
@@ -0,0 +1,136 @@
...
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - VMLiveUpdateFeatures
+```
+
+### Configure the Workload Update Strategy
+
+Configure `LiveMigrate` as `workloadUpdateStrategy` since the current implementation of the hotplug process requires the VM to live-migrate.
```
aburdenthehand, line 23: This might seem obvious but could we also mention that this action is performed in the KubeVirt CR as well

docs/operations/memory_hotplug.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+Since the Virtual Machine is now running we can patch the VM object to double the available guest memory so that we'll go from 128Mi to 256Mi.
+
+```sh
+$ kubectl patch vm vm-cirros -p='[{"op": "replace", "path": "/spec/template/spec/domain/memory/guest", "value": "256Mi"}]' --type='json'
+```
+
+After the hotplug request is processed and the Virtual Machine is live migrated the new amount of memory should be available to the guest
```
aburdenthehand, line 120:
```suggestion
After the hotplug request is processed and the Virtual Machine is live migrated, the new amount of memory should be available to the guest
```

docs/operations/memory_hotplug.md
```diff
@@ -0,0 +1,136 @@
...
+```
+```json
+{
+  "guestAtBoot": "128Mi",
+  "guestCurrent": "256Mi",
+  "guestRequested": "256Mi"
+}
+```
+
+## Limitations
```
aburdenthehand, line 134: Would it be better to include the limitations in the introduction instead of at the end of the tutorial?

aburdenthehand (commented):
Great work @acardace ! I particularly like the inclusion of the tutorial to show the full process together.
I just have a few comments/suggestions.

docs/operations/memory_hotplug.md
acardace: It's a good suggestion but honestly, I can't think of anything else to add, memory hotplug is in general a well-understood concept in virtualization.

docs/operations/memory_hotplug.md
acardace: yes, that's a good point.

acardace:
@aburdenthehand can you please take another look? I think I addressed all your comments.

docs/operations/memory_hotplug.md
```diff
@@ -0,0 +1,136 @@
...
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+spec:
+  liveUpdateFeatures:
+    memory:
+      maxGuest: 2Gi
+```
+Specifying the memory.maxGuest value is optional. If left unset, it defaults to the value defined in maxHotplugRatio inside the KubeVirt CR (refer to [KubeVirt API Reference](https://kubevirt.io/api-reference/v1.1.0/definitions.html#_v1_liveupdateconfiguration)). For example, if a VM is configured with 512Mi of guest memory and maxGuest is not defined, and maxHotplugRatio is 2, then maxGuest will equal 1Gi.
+
+**NOTE:** If memory hotplug is enabled on an already running VM a reboot is necessary for the changes to take effect.
```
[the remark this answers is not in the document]
acardace: I changed it to enabled/disabled.

PR #757, opened 2024-01-23
Move mazzystr to emeritus

dhiller:
@mazzystr has left the project a while ago, so we are moving him to emeritus.

/cc @aburdenthehand


aburdenthehand:
/lgtm
@mazzystr We thank you for your service!

dhiller:
/approve

Self-approving since there's no harm to be caused.

PR #764, opened 2024-02-05
networking,bridge: Mitigate network disruption during live migration

ormergi:
Update user guide on how to mitigate network disruption during live migration

Fixes #

**Checklist**

**Release note**:

```release-note

```

### DIFF
--- a/docs/virtual_machines/interfaces_and_networks.md
+++ b/docs/virtual_machines/interfaces_and_networks.md
@@ -490,6 +490,14 @@ spec:
 > default interface type to `masquerade`, and disabling the `bridge`
 > type for pod network, as shown in the example above.
 
+> **Note:** Migrating a VM with secondary interfaces that uses bridge 
+> binding may cause long periods of traffic disruption. 
+> One of the reasons is that the migration destination pod interface advertises 
+> about its existence before the migration is completed.
+> If Bridge CNI is used, it's possible to avoid the issue by setting the
+> `disableContainerInterface` option as follows: 
+> https://www.cni.dev/plugins/current/main/bridge/#example-l2-only-disabled-interface-configuration
+
 ### slirp
 
 In `slirp` mode, virtual machines are connected to the network backend


docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -490,6 +490,14 @@ spec:
 > default interface type to `masquerade`, and disabling the `bridge`
 > type for pod network, as shown in the example above.
 
+> **Note:** Migrating a VM with secondary interfaces that uses bridge 
+> binding may cause long periods of traffic disruption. This occurs when
+> the interface is defined with an explicit MAC address (manually or
```
AlonaKaplan, line 495: s/explicit/custom

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -490,6 +490,14 @@ spec:
 > default interface type to `masquerade`, and disabling the `bridge`
 > type for pod network, as shown in the example above.
 
+> **Note:** Migrating a VM with secondary interfaces that uses bridge 
+> binding may cause long periods of traffic disruption. This occurs when
+> the interface is defined with an explicit MAC address (manually or 
+> automatically through KubeMacPool) on nodes that have IPv6 enabled.
```
AlonaKaplan, line 496: I wouldn't mention the low level details about the MAC and the IPv6. 
I would just say - "One of the reasons is that the target pod interface announces about its existence before the migration is completed."

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -490,6 +490,14 @@ spec:
...
+> In order to mitigate traffic disruption during live migration consider
```
AlonaKaplan, line 497: "If bridge CNI is used, it is possible to avoid this issue by setting..."

ormergi: Note that this note is under `bridge` section

docs/virtual_machines/interfaces_and_networks.md
ormergi: Setting custom MAC and having IPv6 enabled is the terms for this issue in the first place.
I can add what you suggest,  but I think its worth mentioning the terms, WDYT?

docs/virtual_machines/interfaces_and_networks.md
AlonaKaplan: under bridge binding section. Bridge ibnding can be used with various CNIs (bridge-cni, ovs-cni, ovn-kube, etc...)

docs/virtual_machines/interfaces_and_networks.md
AlonaKaplan: I think that those are low lover implementation details that may confuse the reader.

ormergi: Done

docs/virtual_machines/interfaces_and_networks.md
ormergi: Done

docs/virtual_machines/interfaces_and_networks.md
```diff
@@ -490,6 +490,14 @@ spec:
 > default interface type to `masquerade`, and disabling the `bridge`
 > type for pod network, as shown in the example above.
 
+> **Note:** Migrating a VM with secondary interfaces that uses bridge 
+> binding may cause long periods of traffic disruption. 
+> One of the reasons is that the migration destination pod interface advertises 
+> about its existence before the migration is completed.
+> If Bridge CNI is used, it's possible to avoid the issue by setting the
+> `disableContainerInterface` option as follows:
```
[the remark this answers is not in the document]
phoracek: I agree that having this in a note tied to the binding is a problem. +1 on not keeping this here.

However, we want to make the user-guide user-friendly and copy-pasteable. IIUIC there is no harm in setting this new attribute by default for all NADs that are to be used by a VM.

 Although the current set of examples is all over the place (using OVS CNI, bridge, cnv-bridge, ptp)... Please update the places where we use NetworkAttachmentDefinition cnv-bridge or bridge, so they always set this new attribute.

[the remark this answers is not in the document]
phoracek: ... since we can't use this attribute with IPAM, let's just use it in the macspoof example that currently uses `cnv-bridge`. And we can rename it to `bridge` while at it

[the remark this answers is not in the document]
ormergi: If a user tackle this issue with packet drops during migraion how will they know how to handle it? its not that the bug is referenced anywhere.
Having this note in the user-guide could save some time digging and frustration for users who actually tackle it.
There are some conditions for this issue to accrue, we can say its a known issue with a work around.
WDYT?

Regarding the NAD examples, there are two example that spesify the bridge CNI:
- [IPAM example](https://github.com/kubevirt/user-guide/blob/main/docs/virtual_machines/interfaces_and_networks.md?plain=1#L147), 
The new `disableContainerInterface` cannot work with IPAM settings.
- [Macspoof example](https://github.com/kubevirt/user-guide/blob/main/docs/virtual_machines/interfaces_and_networks.md?plain=1#L1092), I honestly dont think we should piggy back this example, it may be confusing to see it related to macspoof, and make one they do related while they dont.

Having and example for the new option will be the most copy-past friendly IMP.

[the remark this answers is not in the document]
EdDev: I do not think it makes sense to document every bug we resolve through integration settings.
The solution is to set a CNI bridge with the mentioned config, but KubeVirt is trying to to bind itself to a specific CNI and therefore this problematic.

Users that may have this problem can open a support ticket, search the web or ask the community through other means.

I do not mind if this is not mentioned at all or through an existing example.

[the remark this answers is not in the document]
AlonaKaplan: We can change the ovs example [here](https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#multus) to bridge.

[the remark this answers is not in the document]
phoracek: > it may be confusing to see it related to macspoof, and make one they do related while they dont.

It won't be confusing if we change the ovs example as well. Then the only thing extra in macspoof example is macspoof setting.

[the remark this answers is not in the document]
ormergi: OK I will post another PR to change the ovs example.
Thanks for the discussion.

ormergi:
Closing this PR in favor of #765 

/close

PR #772, opened 2024-03-12
persistent reservation: add security note

alicefr:
Add details that the persistent reservation deploys a privileged component and enables additional privileged operations.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/virtual_machines/disks_and_volumes.md
+++ b/docs/virtual_machines/disks_and_volumes.md
@@ -74,6 +74,11 @@ configuration:
     -  PersistentReservation
 ```
 
+> **Note:** The persistent reservation feature enables an additional privileged
+> component to be deployed together with virt-handler. Because this feature allows
+> for sensitive security procedures, it is disabled by default and requires cluster
+> administrator configuration.
+
 ### disk
 
 A `disk` disk will expose the volume as an ordinary disk to the VM.


docs/virtual_machines/disks_and_volumes.md
```diff
@@ -74,6 +74,11 @@ configuration:
     -  PersistentReservation
 ```
 
+The persistent reservation feature enables an additional privileged component to
```
fabiand, line 77: What about the following markup:

> [!WARNING]
> warn text

fabiand: This makes me wonder: How can an admin limit the use of this feature inside a cluster? aka limit it to a usr or to certain vms?

generally speaking something like kyverno could be used.

alicefr: Either you have persistent reservation enable at cluster level or not. My suggestion is to have it disable is because it deploys an additional privileged component, and this feature is quite niece. I wouldn't enable always by default.
In order to use SCSI persistent reservation, you also need to have a CSI plugin which support SCSI protocol. This feature requires in any case some work from the cluster-adin.

However, from regular user, if this is enable, I don't think we should limit this only to certain namespaces.

I hope this makes the entire feature a bit clearer. Please let me know, which aspect you would like me to capture in the documentation.

aburdenthehand: An [admonition without a title](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#removing-the-title) (or a note admonition tbh) might be useful to bring attention to this para. It does seem worthy, and in its position it is a bit easy to miss. 
```
!!! note ""

    yada yada
```

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -74,6 +74,11 @@ configuration:
...
+be deployed together with virt-handler. Because this feature allows for sensitive
+security procedures, it is by default disabled and requires cluster administrator
```
aburdenthehand, line 79:
```suggestion
security procedures, it is disabled by default and requires cluster administrator
```

aburdenthehand (commented):
Two nits but either way this is a good addition.

alicefr:
@aburdenthehand @fabiand could you please check this documentation once more?

docs/virtual_machines/disks_and_volumes.md
aburdenthehand: @alicefr Sorry to be a pain but the note text needs 4 spaces of indentation to render. 
It's currently rendering a single line: https://deploy-preview-772--kubevirt-user-guide.netlify.app/virtual_machines/disks_and_volumes/#persistent-reservation

alicefr: @aburdenthehand sorry, I used the > as we already have in the documentation

aburdenthehand:
/lgtm
@alicefr Totally fair :) Thank you!

aburdenthehand:
@phoracek

phoracek (commented):
/approve

PR #792, opened 2024-04-29
document an example how to use a static PersistentVolume

aep:
thanks to @alicefr for helping me figure this out.
maybe the example is useful to someone else.

### DIFF
--- a/docs/virtual_machines/disks_and_volumes.md
+++ b/docs/virtual_machines/disks_and_volumes.md
@@ -483,6 +483,61 @@ Enabling this feature does two things:
 - If the disk is a Filesystem PVC, the matching file is expanded
   to the remaining size (while reserving some space for file system overhead).
 
+#### Statically provisioned block PVCs
+
+To use an externally managed local block device from a host ( e.g. /dev/sdb , zvol, LVM, etc... ) in a VM directly,
+you would need a provisioner that supports block devices, such as OpenEBS LocalPV.
+
+Alternatively, local volumes can be provisioned by hand.
+I.e. the following PVC:
+
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: myblock
+spec:
+  storageClassName: local-device
+  volumeMode: Block
+  accessModes:
+    - ReadWriteOnce
+  resources:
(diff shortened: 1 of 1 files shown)


docs/virtual_machines/disks_and_volumes.md
```diff
@@ -412,6 +412,56 @@ spec:
         claimName: mypvc
 ```
 
+
+#### block device from host
```
alicefr, line 416: I would call this `Statically provisioned block PVCs`

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -412,6 +412,56 @@ spec:
...
+
```
alicefr, line 417: Here, are missing a short introduction why we need statically PVC. You could mentioned, your use-case to share a disk from the host to the VM

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -412,6 +412,56 @@ spec:
...
+        - key: kubernetes.io/hostname
+          operator: In
+          values:
+          - my-node
+  accessModes:
+    - ReadWriteOnce
+  capacity:
+    storage: 100Gi
+  local:
+    path: /dev/sdb
```
alicefr, line 462: Please, split the storage class and PV creation to the PVCs definition. We should make clear here that PVs ans SCs can only be created by the cluster-admin and not by regular users. Regular users can only get storage classes and create PVCs

docs/virtual_machines/disks_and_volumes.md
```diff
@@ -412,6 +412,62 @@ spec:
...
+  resources:
+    requests:
+      storage: 100Gi
+```
+
+can claim a PersistentVolume pre-created by a cluster admin like so:
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
```
mhenriks, line 443: I haven't tried this in awhile but am surprised that the `StorageClass` resource is necessary.  What happens if it is omitted?

aep: it'll say it doesnt know the storageclass

PR #802, opened 2024-05-28
Workaround: adding mkdocs-redirects build and redirect map to mkdocs.yml

aburdenthehand:
The redirects for #801 are not working due to something jekyll related on the website or automation end.
Until that is figured out, this is an alternative method of redirecting our guide, using mkdocs-redirect.

mkdocs-redirects requires the redirect map to be in the mkdocs.yml file, and adds mkdocs-redirects to Makefile and netlify.toml
Unrelated, but this PR also removes an unnecessary copy of the netlify _redirects file in the root directory (it is in the docs/ dir). 

This has been tested locally. 
Note: The netlify preview build here should not be used to test as it picks up the original _redirect file from #801, which works fine in the preview build, but is being removed somewhere between netlify and the web hosting. 

There is a sibling PR in project-infra to add mkdocs-redirects to the user-guide image: https://github.com/kubevirt/project-infra/pull/3442

**Release note**:

```release-note
NONE
```


dhiller (approved):
/approve

aburdenthehand:
Thanks @dhiller !
I can confirm the redirects are now working live as expected. Seems as though the PR to project-infra (#3442) is not needed after all.

PR #805, opened 2024-06-06
Document update volume strategy and volume migration

alicefr:
Document update volume strategy and volume migration

**Release note**:

```release-note
Add documentation for the update volume strategy and volume migration
```

### DIFF
--- a/docs/storage/volume_migration.md
+++ b/docs/storage/volume_migration.md
@@ -1,0 +1,214 @@
+# Migration update volume strategy and volume migration
+
+Storage migration is possible while the VM is running by using the update volume strategy. Storage migration can be useful in the cases where the users need to change the underlying storage, for example, if the storage class has been deprecated, or there is a new more performant driver available.
+
+This feature doesn't handle the volume creation or cover migration between storage classes, but rather implements a basic API which can be used by overlaying tools to perform more advanced migration planning.
+
+If `Migration` is specified as `updateVolumesStrategy`, KubeVirt will try to migrate the storage from the old volume set to the new one when the VirtualMachine spec is updated. The migration considers the changed volumes present into a single update. A single update may contain modifications to more than one volume, but sequential changes to the volume set will be handled as separate migrations.
+
+Updates are declarative and [GitOps](https://kubevirt.io/user-guide/cluster_admin/gitops/) compatible. For example, a new version of the VM specification with the new volume set and the migration volume update strategy can be directly applied using `kubectl apply` or interactively editing the VM with `kubectl edit`
+
+Example:
+Original VM with a datavolume and datavolume template:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  labels:
+    kubevirt.io/vm: vm-dv
+  name: vm-dv
+spec:
+  dataVolumeTemplates:
+  - metadata:
+      creationTimestamp: null
+      name: src-dv
+    spec:
+      storage:
+        accessModes:
+        - ReadWriteOnce
+        volumeMode: Filesystem
+        resources:
+          requests:
+            storage: 2Gi
+        storageClassName: local
+      source:
+        registry:
+          url: docker://registry:5000/kubevirt/alpine-container-disk-demo:devel
+  runStrategy: "Always"
+  template:
+    metadata:
+      labels:
+        kubevirt.io/vm: vm-dv
+    spec:
+      domain:
+        devices:
+          disks:
+          - disk:
+              bus: virtio
+            name: datavolumedisk1
+          interfaces:
+          - masquerade: {}
+            name: default
+        resources:
+          requests:
+            memory: 128Mi
+      terminationGracePeriodSeconds: 0
+      volumes:
+      - dataVolume:
+          name: src-dv
+        name: datavolumedisk1
+      networks:
+      - name: default
+        pod: {}
+```
+
+The datavolume `src-dv` is migrated to the pvc `dest-pvc`:
+```yaml
+apiVersion: cdi.kubevirt.io/v1beta1
+kind: DataVolume
+metadata:
+  name: dst-dv
+spec:
+  source:
+      blank: {}
+  storage:
+    accessModes:
+      - ReadWriteMany
+    volumeMode: Block
+    resources:
+      requests:
+        storage: 5Gi
+    storageClassName: rook-ceph
+```
+
+by updating the VM:
+```diff
+ apiVersion: kubevirt.io/v1
+ kind: VirtualMachine
+     kubevirt.io/vm: vm-dv
+   name: vm-dv
+ spec:
++  updateVolumesStrategy: Migration
+   dataVolumeTemplates:
+   - metadata:
+-      name: src-pvc
++      name: dst-dv
+
+       volumes:
+       - dataVolume:
+-          name: src-pvc
++          name: dst-dv
+         name: datavolumedisk1
+```
+
+The destination volume may be of a different type or size than the source. It is possible to migrate from and to a block volume as well as a filesystem volume.
+The destination volume should be equal to or larger than the source volume. However, the additional difference in the size of the destination volume is not instantly visible within the VM and must be manually resized because the guest is unaware of the migration.
+
+The volume migration depends on the `VolumeMigration` and `VolumesUpdateStrategy` feature gates and the `LiveMigrate` workloadUpdateStrategy. To fully enable the feature, add the following to the KubeVirt CR:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - VMLiveUpdateFeatures
+        - VolumesUpdateStrategy
+        - VolumeMigration
+    vmRolloutStrategy: LiveUpdate
+  workloadUpdateStrategy:
+    workloadUpdateMethods:
+    - LiveMigrate
+```
+
+The volume migration progress can be monitored by watching the corresponding VirtualMachineInstanceMigration object using the label `kubevirt.io/volume-update-in-progress: <vm-name>`. Example:
+```bash
+$ kubectl get virtualmachineinstancemigrations -l kubevirt.io/volume-update-in-progress: vmi` --watch=true
+NAME                           PHASE       VMI
+kubevirt-workload-update-abcd  Running     vmi
+```
+
+## Datavolume template
+
+Updating a datavolume that is referenced by a datavolume template requires special caution. The volumes section must include a reference to the name of the datavolume template. This means that the datavolume templates must either be entirely deleted or updated as well.
+
+Example of updating the datavolume for the original VM in the first example:
+```yaml
+apiVersion: cdi.kubevirt.io/v1beta1
+kind: DataVolume
+metadata:
+  name: dst-dv
+spec:
+  source:
+      blank: {}
+  storage:
+    accessModes:
+      - ReadWriteMany
+    volumeMode: Block
+    resources:
+      requests:
+        storage: 10Gi
+    storageClassName: rook-ceph-block
+---
+apiVersion: kubevirt.io/v1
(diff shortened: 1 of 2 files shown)


docs/storage/volume_migration.md
```diff
@@ -0,0 +1,204 @@
...
+
+Only certain types of disks and volumes are supported to be migrated. For an invalid type of volume the RestartRequired condition is set and volumes will be replaced upon VM restart.
+Currently, the volume migration is supported between PersistentVolumeClaims and Datavolumes.
+Additionally, volume migration is forbidden if the volumes is:
+* shareable, since it cannot guarantee the data consistency with multiple writers
+* hotpluggable, this case isn't currently supported
+* filesystem, since virtiofs doesn't curretly support live-migration
+* lun, originally the disk might support SCSI protocol but the destination PVC class does not. This case isn't currently supported.
+
+Currently, KubeVirt only enables live migration between separate nodes. Volume migration relies on live migration; hence, warm migrating storage on the same node is also not possible.
```
awels, line 204: So there are 3 types of migration:
1. Cold. The VM is turned off, and the disks are migrated, then the VM is turned on.
2. Warm. The VM is running, and a snapshot of the disk is copied to the target. Then the VM is turned off and the remaining delta between the source and target is copied, then the VM is turned on again. This should be much faster than a cold migration as there is less data to copy while the VM is turned off.
3. Hot (or live) migration. This what this document is about. So I would change the wording according to the suggestion.
```suggestion
Currently, KubeVirt only enables live migration between separate nodes. Volume migration relies on live migration; hence, live migrating storage on the same node is also not possible.
```

awels (commented):
Nice, just one small change.

docs/storage/volume_migration.md
alicefr: Replaced, thanks!

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,219 @@
...
+spec:
+  source:
+      blank: {}
+  pvc:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 2Gi
+    storageClassName: local
```
jpeimer, line 156: So if in the original VM, DV also uses the `local` storage class, then in `dst-dv` we can use another storage class?

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,219 @@
...
+      name: dst-dv
+    spec:
+      pvc:
+        accessModes:
+        - ReadWriteOnce
+        volumeMode: Filesystem
+        resources:
+          requests:
+            storage: 2Gi
+        storageClassName: local
```
jpeimer, line 177: same question here

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+## Limitations
+
+Only certain types of disks and volumes are supported to be migrated. For an invalid type of volume the RestartRequired condition is set and volumes will be replaced upon VM restart.
+Currently, the volume migration is supported between PersistentVolumeClaims and Datavolumes.
+Additionally, volume migration is forbidden if the volumes is:
+* shareable, since it cannot guarantee the data consistency with multiple writers
+* hotpluggable, this case isn't currently supported
+* filesystem, since virtiofs doesn't curretly support live-migration
```
jpeimer, line 216: So the source or the destination volume can't be filesystem?

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,219 @@
...
+* lun, originally the disk might support SCSI protocol but the destination PVC class does not. This case isn't currently supported.
+
+Currently, KubeVirt only enables live migration between separate nodes. Volume migration relies on live migration; hence, live migrating storage on the same node is also not possible.
```
jpeimer, line 219: So both source and dest volumes should have `ReadWrite**Many**` accessModes?

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,219 @@
...
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - VMLiveUpdateFeatures
+        - VolumesUpdateStrategy
+        - VolumeMigration
```
jpeimer, line 131: Here, there's no `VolumeMigration`, is that correct?
https://github.com/kubevirt/user-guide/pull/805/files#diff-176c70f4872d2227af90873ef4d4bb5cae0f2709836a794179596f27ea11e5bdR2147-R2148

jpeimer (commented):
Thanks for the doc! 
I'm trying to understand the feature better, so left some questions

docs/storage/volume_migration.md
alicefr: Yes, the PVC specification can differ from volume mode, to access mode, to storage class or size

docs/storage/volume_migration.md
alicefr: Storage migration is currently supported for filesystem types

docs/storage/volume_migration.md
alicefr: Yes because the dependency tree is `VolumeMigration` depends on `VolumesUpdateStrategy` which depends on `VMLiveUpdateFeatures`.

alicefr:
@jpeimer I'll fix the example with the options you suggested. Please, let me know if there is still something unclear

docs/storage/volume_migration.md
jpeimer: Oh, okay, so filesystem is not forbidden, and this line can be removed from the list? 

> Additionally, volume migration is forbidden if the volumes is:
> ...
> * filesystem, since virtiofs doesn't curretly support live-migration

docs/storage/volume_migration.md
jpeimer: Got it, thanks

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,219 @@
...
+          name: dst-dv
+        name: datavolumedisk1
+      networks:
+      - name: default
+        pod: {}
+```
+
+## Limitations
+
+Only certain types of disks and volumes are supported to be migrated. For an invalid type of volume the RestartRequired condition is set and volumes will be replaced upon VM restart.
```
jpeimer, line 211: > For an invalid type of volume the RestartRequired condition is set and volumes will be replaced upon VM restart.

So if the storage type is valid - migration will happen right after changing the VM yaml, and the restart is required?

docs/storage/volume_migration.md
jpeimer: Ack, thanks

docs/storage/volume_migration.md
jpeimer: Thanks. Would it be good to use a different sc name in this example?

docs/storage/volume_migration.md
alicefr: Correct, if the update of the volume is valid then the migration is automatically triggered, otherwise, the Restart Required condition is set.

jpeimer: Got it, thanks

docs/storage/volume_migration.md
alicefr: Sorry, no it isn't support, but here we mean the disk type filesystem no the filesystem PVC. Virtiofs doesn't support migration yet

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,219 @@
...
+spec:
+  dataVolumeTemplates:
+  - metadata:
+      creationTimestamp: null
+      name: src-dv
+    spec:
+      pvc:
+        accessModes:
+        - ReadWriteOnce
+        volumeMode: Filesystem
```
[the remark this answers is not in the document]
alicefr: Done

alicefr:
Fixed the latest comments

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,222 @@
+# Migration update volume strategy and volume migration
+
+Storage migration is possible while the VM is running by using the update volume strategy. Storage migration can be useful in the cases where the users need to change the underlying storage, for example, if the storage class has been deprecated, or there is a new more performant driver available.
+
+This feature doesn't handle the volume creation or cover migration between storage classes, but rather implement a basic API which can be used by overlaying tools to perform more advanced migration planning.
```
dhiller, line 5:
```suggestion
This feature doesn't handle the volume creation or cover migration between storage classes, but rather implements a basic API which can be used by overlaying tools to perform more advanced migration planning.
```

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,222 @@
...
+
+If `Migration` is specified as `updateVolumesStrategy`, KubeVirt will try to migrate the storage from the old volume set to the new one when the VirtualMachine spec is updated. The migration considers the changed volumes present into a single update. A single update may contain modifications to more than one volume, but sequential changes to the volume set will be handled as separate migrations.
```
dhiller, line 7: It might make sense to describe how users of this feature can follow the status of the migration?

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,222 @@
...
+
+The updates are declarative and [GitOps](https://kubevirt.io/user-guide/cluster_admin/gitops/) compatible. For example, a new version of the VM specification with the new volume set and the migration volume update strategy can be directly applied using `kubectl apply` or interactively editing the VM with `kubectl edit`
```
dhiller, line 9: nit

```suggestion
Updates are declarative and [GitOps](https://kubevirt.io/user-guide/cluster_admin/gitops/) compatible. For example, a new version of the VM specification with the new volume set and the migration volume update strategy can be directly applied using `kubectl apply` or interactively editing the VM with `kubectl edit`
```

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,222 @@
...
+spec:
+  accessModes:
+  - ReadWriteMany
+  volumeMode: Block
+  resources:
+    requests:
+      storage: 5Gi
+```
+
+by updating the VM:
```
dhiller, line 80: I got confused when reading this since the elements were sorted differently in this version.

Maybe too technical - I'd want to have emphasized what to change, so I can see it better. Maybe use a diff here to show what exactly I need to add and/or change. Or maybe specify in a block what has been added or changed?

docs/storage/volume_migration.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+## Limitations
+
+Only certain types of disks and volumes are supported to be migrated. For an invalid type of volume the RestartRequired condition is set and volumes will be replaced upon VM restart.
+Currently, the volume migration is supported between PersistentVolumeClaims and Datavolumes.
+Additionally, volume migration is forbidden if the disk is:
+* shareable, since it cannot guarantee the data consistency with multiple writers
+* hotpluggable, this case isn't currently supported
+* filesystem, since virtiofs doesn't curretly support live-migration
```
dhiller, line 219:
```suggestion
* filesystem, since virtiofs doesn't currently support live-migration
```

dhiller (commented):
Hey @alicefr , I've added one question inline and some nits, will not block on these, however, will

/hold

to give you the chance to react if you want. Feel free to unhold, or ping me again if you decided to update.

/lgtm

Thank you for your patience with me :blush:

docs/storage/volume_migration.md
alicefr: My idea, here, was that the user could copy the 2 full yamls as example and applied them separately, but you are probably right a diff will be better

docs/storage/volume_migration.md
alicefr: Yes, I have added how to watch the migration using the label

alicefr:
@dhiller I addressed your feedback, many thanks for the review!

dhiller (approved):
Thank you @alicefr !

PR #849, opened 2024-10-30
Change spec.running refs to runStrategy

arnongilboa:
Change deprecated spec.running references to runStrategy.

**Which issue(s) this PR fixes**:
jira-ticket: https://issues.redhat.com/browse/CNV-45433

**Release note**:
```release-note
Change deprecated spec.running references to runStrategy
```

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -160,27 +160,15 @@ with `spec.replica` set to `1`.
 ## How to use a VirtualMachine
 
 A VirtualMachine will make sure that a VirtualMachineInstance object
-with an identical name will be present in the cluster, if `spec.runStrategy`
-is set to `Always`. Further it will make sure that a
-VirtualMachineInstance will be removed from the cluster if
-`spec.runStrategy` is set to `Halted`.
-
-`spec.runStrategy` has several additional options for
-controlling the state of the associated VirtualMachineInstance object.
-An extended explanation of `spec.runStrategy` can be
-found in [Run Strategies](./compute/run_strategies.md)
+with an identical name will be present in the cluster when the VirtualMachine
+is in a Running state, which is controlled via the `spec.runStrategy` field.
+For more information regarding Run Strategies, please refer to [Run Strategies](./compute/run_strategies.md)
 
 ### Starting and stopping
 
-After creating a VirtualMachine it can be switched on or off like this:
-
-    # Start the virtual machine:
-    virtctl start vm
-
-    # Stop the virtual machine:
-    virtctl stop vm
-
-`kubectl` can be used too:
+Virtual Machines can be turned on/off in an imperative or a declarative manner.
+Setting a `spec.runStrategy` like `Always` or `Halted` means that the system will continuously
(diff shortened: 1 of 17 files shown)


docs/architecture.md
```diff
@@ -160,17 +160,14 @@ with `spec.replica` set to `1`.
...
 
-There exists a field `spec.runStrategy` which can also be used to
-control the state of the associated VirtualMachineInstance object. To
-avoid confusing and contradictory states, these fields are mutually
-exclusive.
-
-An extended explanation of `spec.runStrategy` vs `spec.running` can be
+`spec.runStrategy` has several additional options for
+controlling the state of the associated VirtualMachineInstance object.
+An extended explanation of `spec.runStrategy` can be
```
iholder101, line 170: Looks good to me. An alternative to this explanation could be something like:
```
A VirtualMachine will make sure that a VirtualMachineInstance object
with an identical name will be present in the cluster when the VirtualMachine
is in a Running state, which is controlled via the `spec.runStrategy` field.
For more information regarding Run Strategies, please refer to [Run Strategies](./compute/run_strategies.md).
```

The rationale is to avoid diving into the different strategies here, and just outline the main idea and refer to the right place for more info.

WDYT?

docs/architecture.md
```diff
@@ -187,11 +184,11 @@ After creating a VirtualMachine it can be switched on or off like this:
 
     # Start the virtual machine:
     kubectl patch virtualmachine vm --type merge -p \
-        '{"spec":{"running":true}}'
+        '{"spec":{"runStrategy": "Always"}}'
```
iholder101, line 187: (Regarding line 183, `kubectl` can be used too:`, that I can't comment on for some reason):

I wonder if we should mention here the difference between the imperative (i.e. virtctl) and declarative (i.e. run strategies) approaches. Maybe we can add a section that says something like:

```
Virtual Machines can be turned on/off in an imperative or a declarative manner.
Setting a Run Strategy like Always or Halted means that the system will continuously
try to ensure the Virtual Machine is turned on/off. However, with the Manual strategy,
the user would imperatively choose when to turn the VM on or off, without the system
performing any automatic actions.
```

WDYT?

iholder101 (commented):
Looks great to me! Thank you @arnongilboa!
Left two small suggestions, let me know what you think!

/lgtm

iholder101:
Also, can you please also documentation regarding `Once` strategy to `docs/compute/run_strategies.md`?
This can be done in a follow-up if you prefer

docs/architecture.md
arnongilboa: Much better!

docs/architecture.md
arnongilboa: Let's split it around the examples, wdyt?

arnongilboa:
> Also, can you please also documentation regarding `Once` strategy to `docs/compute/run_strategies.md`? This can be done in a follow-up if you prefer

added

iholder101:
Thank you @arnongilboa! Looks great!
/lgtm

PR #855, opened 2024-12-05
Reflect the fact that VMsnapshots are already supported as clone source

iholder101:
Reflect the fact that VMsnapshots are already supported as clone source.

**Release note**:

```release-note
Reflect the fact that VMsnapshots are already supported as clone source
```

### DIFF
--- a/docs/storage/clone_api.md
+++ b/docs/storage/clone_api.md
@@ -1,7 +1,11 @@
 # Clone API
 
-The `clone.kubevirt.io` API Group defines resources for cloning KubeVirt objects. Currently, the only supported cloning
-type is `VirtualMachine`, but more types are planned to be supported in the future (see future roadmap below).
+The `clone.kubevirt.io` API Group defines resources for cloning KubeVirt objects.
+
+This API has two main use-cases:
+* Cloning a Virtual Machine.
+* Creating "golden snapshots" that would be used as a template for creating new VMs
+(see [Using clones as a "golden VM image"](#using-clones-as-a-golden-vm-image) below).
 
 Please bear in mind that the clone API is in version `v1alpha1`. This means that this API is not fully stable
 yet and that APIs may change in the future.
@@ -11,7 +15,7 @@ yet and that APIs may change in the future.
 ### Snapshot / Restore
 
 Under the hood, the clone API relies upon Snapshot & Restore APIs. Therefore, in order to be able to use the clone API,
-please see [Snapshot & Restore prerequisites](../storage//snapshot_restore_api.md#prerequisites).
+please see [Snapshot & Restore prerequisites](../storage/snapshot_restore_api.md#prerequisites).
 
 ### Snapshot Feature Gate
 
@@ -74,7 +78,6 @@ The source and target indicate the source/target API group, kind and name. A few
 
 * Currently, the only supported kinds are `VirtualMachine` (of `kubevirt.io` api group) and `VirtualMachineSnapshot` (
 of `snapshot.kubevirt.io` api group), but more types are expected to be supported in the future.
-See "future roadmap" below for more info.
 
 * The target name is **optional**. If unspecified, the clone controller will generate a name for the target automatically.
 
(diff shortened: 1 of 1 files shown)


docs/storage/clone_api.md
```diff
@@ -155,28 +157,18 @@ After the clone is finished, the target can be inspected:
...
 * Create a VM
 
-* Prepare a "golden VM" environment
-
-  * This can mean different things in different contexts. For example, write files, install applications, apply configurations,
-    or anything else.
+* Prepare a "golden VM" environment. This can mean different things in different contexts.
+For example, write files, install applications, apply configurations, or anything else.
 
 * Snapshot the VM
```
ShellyKa13, line 174: why do you have to delete the vm?

docs/storage/clone_api.md
```diff
@@ -155,28 +157,18 @@ After the clone is finished, the target can be inspected:
...
 **Create a golden image**
 
 * Create a VM
 
-* Prepare a "golden VM" environment
-
-  * This can mean different things in different contexts. For example, write files, install applications, apply configurations,
-    or anything else.
+* Prepare a "golden VM" environment. This can mean different things in different contexts.
+For example, write files, install applications, apply configurations, or anything else.
```
ShellyKa13, line 171: I dont quite understand this examples so maybe worth to add a few words or explanations

docs/storage/clone_api.md
```diff
@@ -155,28 +157,18 @@ After the clone is finished, the target can be inspected:
 kubectl get vm vm-clone-target -o yaml
 ```
 
-## Future roadmap
-
-The clone API is in an early alpha version and may change dramatically. There are many improvements and features
-that are expected to be added, the most significant goals are:
-
-* Add more supported source types like `VirtualMachineInstace` in the future.
-* Add a cross-namespace clone support. This needs to be supported for snapshots / restores first.
```
ShellyKa13, line 164: are this 2 goals not expected to happen anymore?

docs/storage/clone_api.md
```diff
@@ -1,7 +1,10 @@
 # Clone API
 
-The `clone.kubevirt.io` API Group defines resources for cloning KubeVirt objects. Currently, the only supported cloning
-type is `VirtualMachine`, but more types are planned to be supported in the future (see future roadmap below).
+The `clone.kubevirt.io` API Group defines resources for cloning KubeVirt objects.
+
+This API is useful both for cloning Virtual Machines,
+but also for creating "golden snapshots" that would be used as a template for creating new VMs
```
ShellyKa13, line 6: In my opinion not sure 'but' is the best word here..
'but' is more fitting if you use 'not only for....' 'but also'
here you put both for ...., but also
seems a bit odd.

ShellyKa13: Also I think you should distinguish between the possible sources of the clone and the use cases of it, and probably put those in different paragraphs :)

ShellyKa13 (commented):
Hi thanks for the update, some comments inline

docs/storage/clone_api.md
iholder101: I don't :)
Added that it's optional to delete it

docs/storage/clone_api.md
iholder101: I've added a more specific example. WDYT?

docs/storage/clone_api.md
iholder101: > Add more supported source types like `VirtualMachineInstace` in the future.

Not really relevant anymore now that the snapshot source is supported.

> Add a cross-namespace clone support. This needs to be supported for snapshots / restores first.

As written, it has to be supported by snapshot/restore first, and I'm not aware of a concrete plan to support it. Perhaps in the further future.

iholder101:
Thanks @ShellyKa13 for your review!

@jean-edouard do you mind having a look as well?

jean-edouard (commented):
/approve

PR #878, opened 2025-03-10
Ensure "Node CPU allocation ratio" is listed in the ToC

morete:
Last section "Node CPU allocation ratio" is not showing on the  table of contents.

This is a minor 1 char addition to the documentation only.


vladikr:
/approve
Thanks @morete !

PR #880, opened 2025-03-30
Update activating_feature_gates.md

pyaillet:
Hi and thanks for this great project :)

I propose this modification because it took me some time to identify which value I should put on the yaml configuration to activate a specific Feature Gate. I hope this change would help others.

Feel free to apply any modification if needed.

### DIFF
--- a/docs/cluster_admin/activating_feature_gates.md
+++ b/docs/cluster_admin/activating_feature_gates.md
@@ -48,5 +48,5 @@ Change the namespace to suite your installation.
 
 ## List of feature gates
 The list of feature gates (which evolve in time) can be checked directly from
-the [source code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/featuregate/active.go#L22) (use the const string values in the yaml configuration).
+the [source code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/featuregate/active.go) (use the const string values in the yaml configuration).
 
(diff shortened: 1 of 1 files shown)


aburdenthehand:
@pyaillet Welcome, and thank you for the PR! This makes sense to me but I'll pull in someone from SIG-compute side to have a look. 
@iholder101 PTAL

docs/cluster_admin/activating_feature_gates.md
```diff
@@ -48,5 +48,5 @@ Change the namespace to suite your installation.
 
 ## List of feature gates
 The list of feature gates (which evolve in time) can be checked directly from
-the [source code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/feature-gates.go#L26).
+the [source code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/featuregate/active.go#L22) (use the const string values in the yaml configuration).
```
iholder101, line 51: Thank you very much @pyaillet :)

WDYT about removing the line number altogether?
```suggestion
the [source code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/featuregate/active.go) (use the const string values in the yaml configuration).
```

The clarification you've added about using the const string is good enough in my opinion, as the line number will likely change again in the future. What do you think?

pyaillet: Yes, of course!

I just applied the suggestion.

iholder101:
Thanks @pyaillet! and welcome to Kubevirt :)
/lgtm

@aburdenthehand PTAL

aburdenthehand:
/approve
Awesome.

PR #888, opened 2025-04-24
update node-labeller docs to include machine types labels

dasionov:
minor updates to the node-labeller docs in regards to the machine type labels that is being used since https://github.com/kubevirt/kubevirt/pull/13690

**Release note**:
```release-note
update node-labeller docs to include machine type labels
```

### DIFF
--- a/docs/compute/virtual_hardware.md
+++ b/docs/compute/virtual_hardware.md
@@ -136,18 +136,18 @@ spec:
         claimName: myclaim
 ```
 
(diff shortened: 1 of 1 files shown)


dasionov:
/cc @Barakmor1 

is this enough? or should i elaborate more about the fact the node-labeller is a must since the machine-type is automatically included in the node-selector

Barakmor1:
/lgtm

Thank you!

PR #890, opened 2025-04-30
chore(doc): document missing features in vmsnapshotrestores

SkalaNetworks:

There's missing features in the documentation of VMRestores:
- target readiness policies
- patches

This was brought to my attention in this issue https://github.com/kubevirt/kubevirt/issues/14588 by @iholder101 

Now every feature available is documented:

```go
// VirtualMachineRestoreSpec is the spec for a VirtualMachineRestoreresource
type VirtualMachineRestoreSpec struct {
	// initially only VirtualMachine type supported
	Target corev1.TypedLocalObjectReference `json:"target"`

	VirtualMachineSnapshotName string `json:"virtualMachineSnapshotName"`

	// +optional
	TargetReadinessPolicy *TargetReadinessPolicy `json:"targetReadinessPolicy,omitempty"`

	// If the target for the restore does not exist, it will be created. Patches holds JSON patches that would be
	// applied to the target manifest before it's created. Patches should fit the target's Kind.
	//
	// Example for a patch: {"op": "replace", "path": "/metadata/name", "value": "new-vm-name"}
	//
	// +optional
	// +listType=atomic
	Patches []string `json:"patches,omitempty"`
}
```

**Checklist**

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/storage/snapshot_restore_api.md
+++ b/docs/storage/snapshot_restore_api.md
@@ -106,6 +106,41 @@ To wait for a restore to complete, execute:
 kubectl wait vmrestore restore-larry --for condition=Ready
 ```
 
+## Target readiness policies
+
+When restoring a `VirtualMachineSnapshot` on an already existing target (like the parent VM of the snapshot), a readiness policy can be specified to adjust how the restore happens if the target VM is not ready. The target VM is considered ready when it is fully stopped. The policy is controlled by setting `targetReadinessPolicy` to one of the available types.
+
+The following policies are available:
+- **WaitGracePeriod** (default policy): Wait 5 minutes for the target VM to be ready. If not ready in time, the restore will fail.
+- **StopTarget**: Stop the target VM so that the restore can continue immediately.
+- **FailImmediate**: Don't wait for the target to be ready before trying to restore. If it is not ready, the restore fails immediately.
+- **WaitEventually**: Kubevirt keeps the `VirtualMachineRestore` around until the target is ready. The restore is started as soon as the target is ready.
+
+## JSON patches
+
+It is sometimes necessary to modify the specs of VMs before restoring them (e.g modifying or adding annotations for CNIs).
+
+JSON patches can be applied to the `specs` and `labels/annotations` of restored VMs using the `patches` parameter:
+
+```yaml
+apiVersion: snapshot.kubevirt.io/v1beta1
+kind: VirtualMachineRestore
+metadata:
+  name: restore-larry
+spec:
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: larry
+  virtualMachineSnapshotName: snap-larry
+  patches:
+    - "{\"op\": \"replace\", \"path\": \"/spec/template/metadata/labels/example\", \"value\": \"replaced-value\"}" # Replace the content of a labels
+```
+
+Patches are only applied if the target is not an already existing VM.
+
+**Keep in mind that patches must be carefully applied.** Some fields might be used to reference other resources, or be used as targets by other resources.
+
 ## Cleanup
 
 Keep `VirtualMachineSnapshots` (and their corresponding `VirtualMachineSnapshotContents`) around as long as you may want to restore from them again.
--- a/docs/storage/snapshot_restore_api.md
+++ b/docs/storage/snapshot_restore_api.md
@@ -116,31 +116,6 @@ The following policies are available:
 - **FailImmediate**: Don't wait for the target to be ready before trying to restore. If it is not ready, the restore fails immediately.
 - **WaitEventually**: Kubevirt keeps the `VirtualMachineRestore` around until the target is ready. The restore is started as soon as the target is ready.
 
-## JSON patches
-
-It is sometimes necessary to modify the specs of VMs before restoring them (e.g modifying or adding annotations for CNIs).
-
-JSON patches can be applied to the `specs` and `labels/annotations` of restored VMs using the `patches` parameter:
-
-```yaml
-apiVersion: snapshot.kubevirt.io/v1beta1
-kind: VirtualMachineRestore
-metadata:
-  name: restore-larry
-spec:
-  target:
-    apiGroup: kubevirt.io
-    kind: VirtualMachine
-    name: larry
-  virtualMachineSnapshotName: snap-larry
-  patches:
-    - "{\"op\": \"replace\", \"path\": \"/spec/template/metadata/labels/example\", \"value\": \"replaced-value\"}" # Replace the content of a labels
-```
-
-Patches are only applied if the target is not an already existing VM.
-
-**Keep in mind that patches must be carefully applied.** Some fields might be used to reference other resources, or be used as targets by other resources.
-
 ## Cleanup
 
 Keep `VirtualMachineSnapshots` (and their corresponding `VirtualMachineSnapshotContents`) around as long as you may want to restore from them again.
--- a/docs/storage/clone_api.md
+++ b/docs/storage/clone_api.md
@@ -127,6 +127,35 @@ This field is used to explicitly set an SMBios serial for the target.
 
 This field is optional. By default, the target would have an auto-generated serial that's based on the VM name.
 
+### JSON patches
+
+It is sometimes necessary to modify the specs of VMs further than filtering before cloning them (e.g modifying or adding annotations for CNIs).
+
+JSON patches can be applied to the `specs` and `labels/annotations` of cloned VMs using the `patches` parameter:
+
+```yaml
+apiVersion: snapshot.kubevirt.io/v1beta1
+kind: VirtualMachineClone
+metadata:
+  name: cone
+spec:
+  source:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: vm-cirros
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: vm-clone-target
+  patches:
+    - '{"op": "add", "path": "/metadata/labels/test", "value": "something"}'
+    - '{"op": "replace", "path": "/spec/template/metadata/labels/example", "value": "replaced-value"}'
+```
+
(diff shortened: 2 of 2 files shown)


docs/storage/snapshot_restore_api.md
```diff
@@ -106,6 +106,41 @@ To wait for a restore to complete, execute:
...
+metadata:
+  name: restore-larry
+spec:
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: larry
+  virtualMachineSnapshotName: snap-larry
+  patches:
+    - "{\"op\": \"replace\", \"path\": \"/metadata/name\", \"value\": \"new-vm-name\"}" # Change the name of the restored VM
```
iholder101, line 137: It's not your fault at all, but this is a bad example that's taken from our docs that is actually invalid since the patches may only change the vm's spec. This was recently fixed in https://github.com/kubevirt/kubevirt/pull/14542, where we removed this misleading example.

Can you please use a different example that modifies spec?

iholder101: As a side-note, we may want to release this validation to also allow patching some fields under `/metadata`. If that's valuable I'm open to discussing it.

SkalaNetworks: Done. I think we should discuss allowing all `/metadata` to be edited as well. There might be some valuable annotations or labels that need to be changed. I know I put labels on VMs with unique identifiers that might need to be changed on restore.

SkalaNetworks: @iholder101 actually the code here https://github.com/kubevirt/kubevirt/blob/9d57a8da9faf79c9699a85a52e2b45e6cfdfedc6/pkg/storage/admitters/vmrestore.go#L218 allows to modify everything under /metadata/labels /metadata/annotations and /spec

docs/storage/snapshot_restore_api.md
```diff
@@ -106,6 +106,41 @@ To wait for a restore to complete, execute:
...
+- **StopTarget**: Stop the target VM so that the restore can continue immediately.
+- **WaitGracePeriod**: Wait 5 minutes for the target VM to be ready. If not ready in time, the restore will fail.
+- **FailImmediate**: Don't wait for the target to be ready before trying to restore. If it is not ready, the restore fails immediately.
+- **WaitEventually**: Kubevirt keeps the `VirtualMachineRestore` around until the target is ready. The restore is started as soon as the target is ready.
+
+## JSON patches
+
+It is sometimes necessary to modify the specs of VMs before restoring them (e.g modifying or adding annotations for CNIs).
+
+JSON patches can be applied to the `specs` of restored VMs using the `patches` parameter:
```
iholder101, line 123: As you've rightfully [mentioned](https://github.com/kubevirt/user-guide/pull/890#discussion_r2068517702), labels and annotations can also be modified. Can you please add that?

docs/storage/snapshot_restore_api.md
```diff
@@ -106,6 +106,41 @@ To wait for a restore to complete, execute:
...
+    kind: VirtualMachine
+    name: larry
+  virtualMachineSnapshotName: snap-larry
+  patches:
+    - "{\"op\": \"replace\", \"path\": \"/spec/template/metadata/labels/example\", \"value\": \"replaced-value\"}" # Replace the content of a labels
+```
+
+Patches are only applied if the target is not an already existing VM.
+
+**Keep in mind that patches must be carefully applied.** Some fields might be used to reference other resources, or be used as targets by other resources.
```
iholder101, line 142: Thinking about this again, won't it make much more sense to document it for clones once your [PR](https://github.com/kubevirt/kubevirt/pull/14617) gets in?

I mean, using the restore object to perform a clone is not something I'd recommend to anyone TBH.
WDYT about waiting with this documentation, then adding it to the clone object instead of here?

docs/storage/snapshot_restore_api.md
```diff
@@ -106,6 +106,41 @@ To wait for a restore to complete, execute:
 kubectl wait vmrestore restore-larry --for condition=Ready
 ```
 
+## Target readiness policies
+
+When restoring a `VirtualMachineSnapshot` on an already existing target (like the parent VM of the snapshot), a readiness policy can be specified to adjust how the restore happens if the target VM is not ready. The target VM is considered ready when it is fully stopped. The policy is controlled by setting `targetReadinessPolicy` to one of the available types.
+
+The following policies are available:
+- **StopTarget**: Stop the target VM so that the restore can continue immediately.
+- **WaitGracePeriod**: Wait 5 minutes for the target VM to be ready. If not ready in time, the restore will fail.
```
ShellyKa13, line 115: You should probably add that this is the default behavior if targetReadinessPolicy was not selected

docs/storage/snapshot_restore_api.md
ShellyKa13: Yeah I agree since patches are valid only for restores to a new VM i.e clone it is better to document the patches in the clone API, restoring to a new VM not through clone is not recommended and actually should not a formal API

ShellyKa13: So you do clone the VM basically.. why cant it be done in the clone API? cause the MAC and IP have to be changed with clone?? If so I think thats where it needs to be changed and make it work the clone to an identical VM except the name.

iholder101: @SkalaNetworks I haven't tested this, but you can configure the clone to use the source's VM mac address. This should work as long as the source VM is shut down, like you've mentioned.

SkalaNetworks: Is this true for the SMBios? It seems to be reset if unspecified.

docs/storage/snapshot_restore_api.md
SkalaNetworks: Done

docs/storage/snapshot_restore_api.md
SkalaNetworks: @iholder101 I still think this feature should be documented, it is a public field available in the CRD, I can make the warning stronger to signify vmrestores should not be used as clones.

But in the end, this is still something that might end up desirable for some end users.

iholder101: > But in the end, this is still something that might end up desirable for some end users.

I just don't see the use-case.
If we can't come up with such a use-case, I'd rather point users to the right place, which is the clone API. The patches field is defined as public mainly for the clone controller to use. Since it is public, theoretically it can be used by users, but again if we can't come up with a sensible use-case I think it'll just be confusing to document that.

iholder101: > Is this true for the SMBios? It seems to be reset if unspecified.

Well, it can be specified :)
Resetting it by default sounds very sensible to me, don't you think?

SkalaNetworks: @iholder101 fair enough. I'll edit this PR to move everything related to Patches in the clone section!

Concenring SMBios, it's sensible to reset it for sure, but only in "pure clones". This is not exaclty the same as my use case which is to restore a snapshot pretty much identically (mac/firmware change) to test a snapshot while keeping the original VM shutdown.

I kinda get confused why the Snapshot API allows for restoring in a separate VM, as it doesn't seem like this feature is recommended? 

Quoting @ShellyKa13 

> Yeah I agree since patches are valid only for restores to a new VM i.e clone it is better to document the patches in the clone API, restoring to a new VM not through clone is not recommended and actually should not a formal API

This is why I initally said the Clone API/Snapshot API had a lot of overlap to my eyes. They can do the exact same stuff in the end, just with different YAML structures and varying amounts of work. I think the doc should be clearer on what the maintainers believe is the best usage of each.

ShellyKa13: Clone API adjusted restore API to support restoring to a new VM with the intention of using this adjusted API only from within the clone API..
Clone API uses this as the best usage the maintainers believe is best.
I think adding to the documentation that restoring the a different VM then the snapshotVMsource should be done by using the Clone API having a vmsnapshot as the source as this is the intended way which takes care for everything required in the process.
@iholder101 WDYT?

iholder101: > Concenring SMBios, it's sensible to reset it for sure, but only in "pure clones". This is not exaclty the same as my use case which is to restore a snapshot pretty much identically (mac/firmware change) to test a snapshot while keeping the original VM shutdown.

@SkalaNetworks right, and "pure clone" is the classic most common use-case, therefore the current defaults make sense to me. You can configure clones for your use-case using the existing configs. I'm not sure what's the problem here tbh :)

In any case, presenting clone-level configs for your use-case (in order to not rely on patches) could be considered.

> This is why I initally said the Clone API/Snapshot API had a lot of overlap to my eyes. 

This is similar to asking "why do we need podman? the user can just configure cgroups, namespaces, SE-linux.. there's an overlap between podman and these technologies".

Sure, you can configure everything yourself. And even podman/docker hand out configurations to directly control cgroups or namespace configurations.

Essentially these are just different levels of abstraction.
If you want to create a container, podman will do many things for you, like validations, optimizations, etc etc. You can always "bypass" it and configure low-level technologies yourself, but then the risk is on you.

I think this is a good analogy for clones and snapshots/restores.

> I think the doc should be clearer on what the maintainers believe is the best usage of each.

Want to restore a VM? use the restore API.
Want to clone a VM? use the clone API.

In your use-case you want to clone a VM, but to configure it to keep its MAC address, etc.
In general - if a new VM is being created - you better use the clone API.

In my eyes this is pretty straight forward and can be understood from the name. But if you wish, we can throw a word in the docs.

SkalaNetworks: Alright, you got me convinced :)

SkalaNetworks:
@iholder101 My PR was merged on Kubevirt, so I moved everything related to patches over to the Clone API

iholder101:
Thank you @SkalaNetworks!
/lgtm

/cc @jean-edouard @aburdenthehand

jean-edouard (commented):
/approve

PR #894, opened 2025-05-09
Document current limitation of dedicatedIOThread on scsi disks

alicefr:
Document current limitation when `dedicatedIOThreads` field is specified on scsi devices. This behavior has been introduced by https://github.com/kubevirt/kubevirt/pull/11219 

**Release note**:

```release-note
NONE
```


docs/storage/disks_and_volumes.md
```diff
@@ -1629,6 +1629,13 @@ Disks will be assigned to IOThreads like this:
     emptydisk5:   2
     emptydisk6:   1
 
+**Note:** `dedicatedIOThreads` isn't currently supported for disks using `scsi` as bus type.
+If the field is set, then the webhook will complain with this error message:
+`IOThreads are not supported for disks on a %s bus`. The reason for this limitation
+consists that KubeVirt supports a single scsi controller for all the disks using the
+`scsi` bus, and QEMU does not support assigning each IOThreads to each scsi disks,
+instead they can only be assigned at the controller level.
```
awels, line 1637: Any mechanism for setting it at the controller level available right now?

alicefr: Since we refuse at the webhook level when we set the dedicatedIOThread, no. The current code could assign a single iothread, like the emulatorThread. However, libvirt does support multiple iothreads for the controller, so we could enhance this mechanism. However, I do belive that this requires some API changes since it doesn't apply on a single disk but at the entire controller level.

PR #993, opened 2026-05-16
Add cloud-init example for guest DNS resolver configuration

ndecandia:

This PR improves the DNS records documentation by adding a practical cloud-init example for configuring the guest DNS resolver.

The current documentation explains the requirement for `options ndots:2` or higher when resolving short names such as `myvmi.mysubdomain`, but it does not provide an example on how to configure the guest operating system.

The added example demonstrates how to configure DNS settings using cloud-init in a persistent and systemd-resolved friendly way.

Fixes #

This is a documentation-only improvement.

**Checklist**

* [x] Design: not required
* [x] PR: The PR description is expressive enough and will help future contributors
* [x] Code: not applicable
* [x] Refactor: not applicable
* [x] Upgrade: not applicable
* [x] Testing: not applicable
* [x] Documentation: user-guide update included
* [x] Community: not required

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/network/dns.md
+++ b/docs/network/dns.md
@@ -86,3 +86,22 @@ like this then:
 > **Note** To resolve short names like `myvmi.mysubdomain` using search domains, the guest's `/etc/resolv.conf` must include `options ndots:2` or higher.
 > In a Linux guest if `options ndots` is not specified, the system defaults to `ndots:1` as indicated in  resolver spec](https://man7.org/linux/man-pages/man5/resolv.conf.5.html).
 > KubeVirt does not configure the guest system resolver. It can either be pre-configured in the VM image or configured at runtime using cloud-init.
+
+Example using cloud-init to configure the guest resolver:
+
+```yaml
+- cloudInitNoCloud:
+    userData: |
+      #cloud-config
+      write_files:
+        - path: /etc/systemd/resolved.conf.d/dns.conf
+          content: |
+            [Resolve]
+            DNS=<cluster-dns-service-ip>
+            Domains=default.svc.cluster.local svc.cluster.local cluster.local
+      runcmd:
+        - systemctl restart systemd-resolved
+  name: cloudinitdisk
+```
+
+The exact DNS service IP depends on the Kubernetes cluster configuration.


jean-edouard (commented):
/approve

docs/network/dns.md
```diff
@@ -85,4 +85,23 @@ like this then:
 
 > **Note** To resolve short names like `myvmi.mysubdomain` using search domains, the guest's `/etc/resolv.conf` must include `options ndots:2` or higher.
 > In a Linux guest if `options ndots` is not specified, the system defaults to `ndots:1` as indicated in  resolver spec](https://man7.org/linux/man-pages/man5/resolv.conf.5.html).
-> KubeVirt does not configure the guest system resolver. It can either be pre-configured in the VM image or configured at runtime using cloud-init.
\ No newline at end of file
+> KubeVirt does not configure the guest system resolver. It can either be pre-configured in the VM image or configured at runtime using cloud-init.
+
+Example using cloud-init to configure the guest resolver:
```
EdDev, line 90: Thanks for adding this example.

A few suggestions:

1. Use networkData for the portable case

The DNS server and search domain configuration can be achieved more portably via networkData (cloud-init's native network config), without assuming systemd-resolved is present or requiring a runcmd service restart. Cloud-init network config (v2) is the canonical mechanism for this and works across distros:

```yaml
- cloudInitNoCloud:
    userData: |
      #cloud-config
    networkData: |
      version: 2
      ethernets:
        eth0:
          dhcp4: true
          nameservers:
            addresses: [<cluster-dns-service-ip>]
            search:
              - default.svc.cluster.local
              - svc.cluster.local
              - cluster.local
  name: cloudinitdisk
```

2. The ndots gap remains unaddressed

The note immediately above the example specifically calls out ndots:2 as the requirement for short-name resolution, but the example does not set it. Note that systemd-resolved does not honor the ndots option — that lives in /etc/resolv.conf. The networkData approach also cannot set ndots, since cloud-init network config v1/v2 has no field for resolver options. If ndots:2 needs to be configured, a userData write_files entry targeting /etc/resolv.conf is required, but this is problematic as well. The example should either address this or clarify that search domains alone are sufficient in most cases.

3. Not in sync with the existing VMI manifest

The document already defines a complete VMI manifest with a cloudinitdisk volume at the top. The new example introduces a floating cloudInitNoCloud snippet that a reader cannot easily relate back to that manifest — it is unclear whether it is a replacement for the existing volume entry or an addition. The example should either show the full VMI manifest with the change applied, or explicitly state that it replaces the cloudinitdisk volume defined earlier.

4. Distro assumption

If the userData / systemd-resolved approach is kept, it should note that this is specific to distributions using systemd-resolved and will not work on RHEL-based systems using NetworkManager, or images using plain /etc/resolv.conf.

EdDev (changes requested):
Thanks for adding this example to our docs.

PR #1001, opened 2026-05-29
Add documentation for when live migration is multithreaded

DorBreger:
this explicitly documents that VMs with explicitly defined cpu limits (not ones created by the autocpulimits feature) will not go through parallel live migration as per [this line of code ](https://github.com/kubevirt/kubevirt/blob/913717faa83bbe8ca6b588c2ee14a5ed465757cf/pkg/virt-handler/migration-source.go#L639).

### DIFF
--- a/docs/compute/live_migration.md
+++ b/docs/compute/live_migration.md
@@ -264,6 +264,7 @@ the VM the guest executes on). There are many reasons for migrations to fail con
 high dirty-rate or low resources like network bandwidth and CPU. On such scenarios, see the following
 alternative strategies below.
 
+For virtual machines that explicitly set resource limits, multithreaded migration is disabled and single-threaded migration will be used instead.
 ### Post-copy
 The way post-copy migrations work is as following:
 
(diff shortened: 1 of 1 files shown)


jean-edouard (commented):
/approve

docs/compute/live_migration.md
```diff
@@ -264,6 +264,7 @@ the VM the guest executes on). There are many reasons for migrations to fail con
 high dirty-rate or low resources like network bandwidth and CPU. On such scenarios, see the following
 alternative strategies below.
 
+For virtual machines that explicitly set resource limits, multithreaded migration is disabled and single-threaded migration will be used instead.
```
jean-edouard:
```suggestion
For virtual machines that explicitly set CPU resource limits, multithreaded migration is disabled and single-threaded migration will be used instead.
```

aburdenthehand:
/lgtm
Thanks for the fix @DorBreger and welcome to KubeVirt!

PR #1004, opened 2026-06-03
missing docs: unprivileged_userfaultfd requirement for postcopy

### LINKED ISSUE #17780: Post-copy runs into permission issues and fix not documented anywhere
**Description**:
Live Migrations using post-copy crash with error:
`internal error: unable to execute QEMU command 'migrate-set-capabilities': Postcopy is not supported: Userfaultfd not available: Operation not permitted`

Searching through the KubeVirt code base I found: https://github.com/kubevirt/kubevirt/pull/8917. So I thought that this issue is caused by the fact CRI-O enforces Seccomp policies that by default prevents access to the `userfaultfd` system call, something that is fundamentally necessary for Postcopy migrations. 

However, even after enabling this featuregate with the relevant profiles I was still unable to Postcopy to work. Then, I found the following OpenShift issue: https://github.com/openshift/machine-config-operator/pull/3724.

Here we find that this permission issue is coming from a level above Seccomp in the kernel. The TLDR is that there are three ways to grant permission to use userfaultfd [[1](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2d5de004e009add27db76c5cdc9f1f7f7dc087e7)]:
1. By allowing unprivileged users to use it explicitly via sysctl
2. By granting CAP_SYS_PTRACE capability to a process / container
3. By exposing the /dev/userfaultfd to a container

OpenShift went with option #1. And I can confirm that running `sysctl -w vm.unpriviledged_userfaultfd=1` does indeed make the above error go away irrespective of whether Seccomp is enabled or not (if someone can clarify why seccomp actually doesn't seem to be required for me that would be helpful).  I also tested option #3 with device plugins and that too works. But which choice is better for KubeVirt?

**What you expected**:
I expect postcopy to work out of the box and at the very least, if some configuration is needed from my end that it should be documented as a pre-req in https://kubevirt.io/user-guide/compute/live_migration/.

**URL**:
https://kubevirt.io/user-guide/compute/live_migration/

Aseeef:

On most modern distros, vm.unprivileged_userfaultfd is disabled by default preventing QEMU from capturing page faults during a postcopy migration on the receiving node. However this requirement is currently not documented anywhere.

Fixes https://github.com/kubevirt/kubevirt/issues/17780

**Checklist**

**Release note**:

```release-note

```

### DIFF
--- a/docs/compute/live_migration.md
+++ b/docs/compute/live_migration.md
@@ -25,6 +25,10 @@ field in the KubeVirt CR must be expanded by adding the `LiveMigration` to it.
 
 - Live migration requires the virt-launcher pod's primary network interface to have the same name on both source and target pods.
 
+- Post-copy live migration requires additional node-level permissions on
+  most environments. See [Node configuration for post-copy](#node-configuration-for-post-copy)
+  for details.
+
 ## Initiate live migration
 
 Live migration is initiated by posting a VirtualMachineInstanceMigration
@@ -223,6 +227,7 @@ To configure `AllowWorkloadDisruption`:
 
 - `AllowWorkloadDisruption` determines whether the migration controller can prioritize completing the migration over avoiding workload disruption.
 - Post-copy migration, when enabled, poses some risk of data loss if a failure occurs during the post-copy phase.
+- Post-copy requires additional node-level permissions on most environments. See [Node configuration for post-copy](#node-configuration-for-post-copy) for details.
 - Pausing the workload facilitates completion of the migration without risk of data loss but may result in temporary workload inactivity.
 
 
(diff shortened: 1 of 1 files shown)


docs/compute/live_migration.md
```diff
@@ -297,6 +302,66 @@ the guest would have to wait for a lot of memory in a short period of time.
...
+  configuration:
+    seccompConfiguration:
+      virtualMachineInstanceProfile:
+        customProfile:
+          localhostProfile: kubevirt/kubevirt.json
+```
+
+##### Mandatory access controls (SELinux / AppArmor)
+
+If the cluster nodes enforce a mandatory access control framework such
```
xpivarc: Here it might be useful to note that it totally depends on the `container_t` and admins might need to configure custom SELinux...

docs/compute/live_migration.md
```diff
@@ -297,6 +302,67 @@ the guest would have to wait for a lot of memory in a short period of time.
...
+runs unprivileged, additional permissions may be required depending on
+your environment.
+
+##### Kernel sysctl
+
+The following sysctl must be enabled on every node that may receive a
+post-copy migration:
+
+```
+vm.unprivileged_userfaultfd=1
```
jean-edouard, line 318: Do you confirm that this is needed even with the seccomp policy deployed? If so, I don't understand how OpenShift cluster successfully run post-copy migrations today...

docs/compute/live_migration.md
```diff
@@ -297,6 +302,67 @@ the guest would have to wait for a lot of memory in a short period of time.
...
+          localhostProfile: kubevirt/kubevirt.json
+```
+
+##### SELinux
+
+On nodes with SELinux enforcing, the virt-launcher process may be
+denied `userfaultfd` depending on the `container_t` policy in use.
+Administrators may need to create a custom SELinux policy module to
+permit this syscall for the virt-launcher context. Consult your
+platform's documentation for the recommended approach.
```
jean-edouard: You can add that nodes that deploy [container-selinux](https://github.com/containers/container-selinux) v2.248+ have the necessary permission (`kernel_userfaultfd_use(container_domain)`).

jean-edouard (commented):
/approve
/hold
Just a couple comments, feel free to unhold.

docs/compute/live_migration.md
Aseeef: Yes, this is needed. OpenShift clusters successfully runs post-copy because https://github.com/openshift/machine-config-operator/pull/3724 modified the machine config operator to set this setting.

docs/compute/live_migration.md
Aseeef: Is the updated paragraph fine?

docs/compute/live_migration.md
jean-edouard: So the instructions below to create the MachineConfig on OpenShift are not needed, are they?

Aseeef: Technically no... Good point, I should update that.

PR #1029, opened 2026-09-02
installation: document host kernel / EL userland compatibility

lyarwood:
**What this PR does**

Adds a short "Host Kernel and Userland Compatibility" subsection to the installation prerequisites.

`virt-launcher` ships an Enterprise Linux (CentOS Stream / RHEL) based libvirt/qemu userland. The host kernel's KVM and vhost interfaces are used to negotiate virtual device features (e.g. virtio-net offloads), so running the EL userland on a substantially different host kernel — particularly a newer, non-EL kernel — can cause feature-negotiation mismatches. These may only surface later, for example as live migration failures where the destination QEMU rejects device state negotiated on the source.

The docs previously described host requirements only in terms of capabilities (`virt-host-validate`, `/dev/kvm`, `/dev/vhost-net`, `/dev/net/tun`) with no guidance on host-kernel/userland alignment.

**Related**

- Tracking issue: kubevirt/kubevirt#19000
- Motivating bug: kubevirt/kubevirt#16386

/sig documentation

### DIFF
--- a/docs/cluster_admin/installation.md
+++ b/docs/cluster_admin/installation.md
@@ -103,6 +103,25 @@ virtualization workloads:
       QEMU: Checking if device /dev/net/tun exists                               : PASS
     ...
 
+### Host Kernel and Userland Compatibility
+
+The `virt-launcher` container ships an Enterprise Linux (CentOS Stream / RHEL)
+based libvirt and QEMU userland. Running this userland on a host kernel that
+diverges significantly from the one it was built against — in particular a
+newer, non-Enterprise-Linux kernel — is not recommended.
+
+The KVM and vhost interfaces exposed by the host kernel are used to negotiate
+virtual device features (for example virtio-net offloads). A host kernel that
+offers a different set of features than the bundled userland expects can lead to
+device feature-negotiation mismatches. These may only surface later, for example
+as live migration failures where the destination QEMU rejects device state that
+was negotiated on the source
+(see [kubevirt/kubevirt#16386](https://github.com/kubevirt/kubevirt/issues/16386)).
+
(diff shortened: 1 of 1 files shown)


0xFelix (commented):
Can you add an overview / matrix, which documents which KubeVirt release images are built on which EL userspace?

nestoracunablanco:
+1 to this PR. Even for an experienced user or developer, finding the root cause could be very difficult. It would be helpful if the KubeVirt Operator had explicit expectations for the host and detected violations during a preflight check, but that is out of scope for this PR.

I’ll be happy to LGTM once @0xFelix’s comment has been addressed.

lyarwood:
> Can you add an overview / matrix, which documents which KubeVirt release images are built on which EL userspace?

@0xFelix Possibly, tables like this do turn into a maintenance burden after a while, for example:

https://docs.openstack.org/nova/latest/reference/libvirt-distro-support-matrix.html

I wonder if we could automate capturing this in the sig-release repo somehow and just link to that from here?

lyarwood:
> > Can you add an overview / matrix, which documents which KubeVirt release images are built on which EL userspace?
> 
> @0xFelix Possibly, tables like this do turn into a maintenance burden after a while, for example:
> 
> https://docs.openstack.org/nova/latest/reference/libvirt-distro-support-matrix.html
> 
> I wonder if we could automate capturing this in the sig-release repo somehow and just link to that from here?

https://github.com/kubevirt/sig-release/issues/69 created to try and automate this

0xFelix (commented):
Sounds reasonable. Would also be great to add a link to the table here later.

/lgtm

aburdenthehand:
/approve
Top stuff

PR #1031, opened 2026-09-14
docs: fix memory hotplug architecture support statement

### LINKED ISSUE #1028: memory hotplug is only supported on AMD64.
**What happened**:
[A clear and concise description of what the bug is](https://kubevirt.io/user-guide/compute/memory_hotplug/) .
The limitations state that memory hotplug is only supported on x86_64 and arm64 architectures, but the code does not support arm64.

pkg/liveupdate/memory.go
`if vmSpec.Architecture != "amd64" {
		return fmt.Errorf("Memory hotplug is only available for x86_64 VMs")
	}`

**What you expected to happen**:
The documentation should be consistent with the code logic.

**How to reproduce it (as minimally and precisely as possible)**:
Steps to reproduce the behavior.

**Additional context**:

**Environment**:
- KubeVirt version (use `virtctl version`): N/A
- Kubernetes version (use `kubectl version`): N/A
- VM or VMI specifications: N/A
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

shashankvarma499:
Fixes #1028.

The limitations list on the memory-hotplug page states that memory
hotplug is supported on both `x86_64` and `arm64` architectures. The
implementation only supports `x86_64`: `pkg/liveupdate/memory/memory.go`
returns "Memory hotplug is only available for x86_64 VMs" for any
`vmSpec.Architecture != "amd64"`.

This updates the user-guide limitations list to match the code, so the
documentation no longer claims arm64 support.


aburdenthehand:
/lgtm
Thanks for picking up this issue @shashankvarma499

